PR #1, opened 2020-08-13
Add the initial ESLint configuration and documentation.

TrevorKarjanis:
- Add a package, tests, and documentation for @ni/eslint-config.

Notes
- The configuration does not include modifications.
- The TypeScript and TSLint support will be added in a seperate PR.

References
- https://eslint.org/docs/developer-guide/shareable-configs
- https://github.com/ni/python-styleguide
- https://github.com/airbnb/javascript
- https://github.com/google/eslint-config-google

Linted and tested
- Verified that the automated test passes
- Verified that the configuration and documentation work using ```npm link```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # NI JavaScript and TypeScript Style Guide
 
+![Logo](resources/logo.svg)
+
 Welcome to NI's JavaScript and TypeScript linter rules for [ESLint](https://eslint.org/docs/user-guide/getting-started) and [TSLint](https://palantir.github.io/tslint/).
 
 ## Installation
--- a/README.md
+++ b/README.md
@@ -1,11 +1,21 @@
-# NI JavaScript Style Guide
+# NI JavaScript and TypeScript Style Guide
 
-Welcome to NI's internal and external JavaScript and TypeScript linter rules for [ESLint](https://eslint.org/docs/user-guide/getting-started) and [TSLint](https://palantir.github.io/tslint/).
+Welcome to NI's JavaScript and TypeScript linter rules for [ESLint](https://eslint.org/docs/user-guide/getting-started) and [TSLint](https://palantir.github.io/tslint/).
 
 ## Installation
 
+Install @ni/eslint-config and its peer dependencies.
+
+Use [npm view](https://docs.npmjs.com/cli/view.html) to get the correct versions of each package.
+
+```bash
+npm view @ni/eslint-config peerDependencies
 ```
-npm install --save-dev eslint eslint-plugin-import @ni/eslint-config
+
+Alternatively, use a shortcut with [install-peerdeps](https://www.npmjs.com/package/install-peerdeps).
+
+```bash
+npx install-peerdeps --dev @ni/eslint-config
 ```
 
 ## Usage
@@ -20,4 +30,4 @@ Extend @ni in the ESLint configuration.
 
 ## License
 
-[MIT (c) 2020 NI](./LICENSE)
+[MIT (c) 2020 National Instruments Corporation](./LICENSE)
--- a/README.md
+++ b/README.md
@@ -1,3 +1,23 @@
 # NI JavaScript Style Guide
 
-Welcome to NI's internal and external JavaScript and TypeScript linter rules for ESLint and TSLint.
+Welcome to NI's internal and external JavaScript and TypeScript linter rules for [ESLint](https://eslint.org/docs/user-guide/getting-started) and [TSLint](https://palantir.github.io/tslint/).
+
+## Installation
+
+```
+npm install --save-dev eslint eslint-plugin-import @ni/eslint-config
+```
+
+## Usage
+
+Extend @ni in the ESLint configuration.
+
+```json
+{
+    "extends": "@ni"
+}
+```
+
+## License
+
+[MIT (c) 2020 NI](./LICENSE)
--- a/README.md
+++ b/README.md
@@ -1,0 +1,3 @@
+# NI JavaScript Style Guide
+
+Welcome to NI's internal and external JavaScript and TypeScript linter rules for ESLint and TSLint.
--- a/package.json
+++ b/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "@ni/javascript-styleguide",
+  "version": "1.0.0",
+  "description": "NI's internal and external JavaScript and TypeScript style guide",
+  "private": true,
+  "scripts": {
+    "eslint-config:install": "cd packages/eslint-config && npm install",
+    "eslint-config:publish": "cd packages/eslint-config && npm publish",
+    "eslint-config:test": "cd packages/eslint-config; npm test",
+    "preinstall": "npm run eslint-config:install",
+    "publish:packages": "npm run eslint-config:publish",
+    "test": "npm run eslint-config:test"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/ni/javascript-styleguide.git"
+  },
+  "author": "NI",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/ni/javascript-styleguide/issues"
+  },
+  "homepage": "https://github.com/ni/javascript-styleguide#readme"
+}
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@ni/javascript-styleguide",
-  "version": "1.0.0",
-  "description": "NI's internal and external JavaScript and TypeScript style guide",
+  "version": "1.0.0-alpha.0",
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -0,0 +1,23 @@
+# NI JavaScript Style Guide
```
jattasNI, line 1: Can we include TypeScript in the heading too? Even though we won't support linting it from day 1, it's good to express our intent for the repo.

README.md
```diff
@@ -0,0 +1,23 @@
...
+## Usage
+
+Extend @ni in the ESLint configuration.
+
+```json
+{
+    "extends": "@ni"
+}
+```
+
```
jattasNI, line 20: At some point (doesn't have to be in this initial submission) we should come up with contributing guidelines. Some ideas of what they'd contain:
 - how do I propose a new rule? who approves it?
 - how do we version this repo? What constitutes a major/minor/patch change?
 - what's our overarching philosophy? why do we want standard style across NI? how do we make decisions about which rules should be active?

I don't necessarily have answers to all of those, just putting it on your radar.

package.json
```diff
@@ -0,0 +1,24 @@
...
+  "name": "@ni/javascript-styleguide",
+  "version": "1.0.0",
+  "description": "NI's internal and external JavaScript and TypeScript style guide",
+  "private": true,
+  "scripts": {
+    "eslint-config:install": "cd packages/eslint-config && npm install",
+    "eslint-config:publish": "cd packages/eslint-config && npm publish",
+    "eslint-config:test": "cd packages/eslint-config; npm test",
+    "preinstall": "npm run eslint-config:install",
+    "publish:packages": "npm run eslint-config:publish",
```
jattasNI, line 11: Have you guys talked about adding a pipeline to do the publish in a future submission? Or would it be a manual process?

README.md
```diff
@@ -0,0 +1,23 @@
...
+## Usage
+
+Extend @ni in the ESLint configuration.
+
+```json
+{
+    "extends": "@ni"
+}
+```
```
jattasNI, line 19: We might also consider some tips about how to go about adopting our rules. Some combination of 
1. describing the commands to run the linter and the fixer
1. describe the workflow of disabling many rules and then enabling rules one-by-one
1. linking to eslint docs about how to configure rules

This could be something we add once we gain more experience in adopting the ruleset though.

package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@ni/javascript-styleguide",
+  "version": "1.0.0",
```
jattasNI, line 3: I've seen some projects start with a 0.*.* version number to indicate the repo is in a prerelease state. Any objection to following that pattern until we've pulled in our first set of rule configurations?

packages/eslint-config/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@ni/eslint-config",
+  "version": "1.0.0",
+  "description": "NI's ESLint Shareable Config",
+  "main": "index.js",
+  "scripts": {
+    "prepublish": "npm test",
```
jattasNI, line 7: I'm curious where you think we'll go with tests in this repo. Right now we're just linting the 3 lines of `index.js`, correct? Do you envision expanding the set of tests? Do you envision running them in a pipeline?

I haven't really thought about what testing looks like for a repo like this, so I'm just curious about what kinds of errors a test suite will be able to catch.

README.md
TrevorKarjanis: You bet!

README.md
TrevorKarjanis: Absolutely, I kept this PR small, functional but basic. Those are things we can tackle next.

package.json
TrevorKarjanis: I think @rajsite  is interested in taking this on. I prefer someone more familiar with the GitHub setup than myself, or someone that will put the time in. I piped it to the root script API at the least. It's based on AirBnB and Google.

README.md
TrevorKarjanis: Sure! 
1. Install
2. Configure
3. Disable
...
Profit!

Naw, seriously all good ideas.

package.json
TrevorKarjanis: Yes, good idea. I usually use npm's semver syntax alpha and beta, because it works with any version.

packages/eslint-config/package.json
TrevorKarjanis: I took a look at both [AirBnB's](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base/test) and [Google's](https://github.com/google/eslint-config-google) testing. AirBnB is testing requires imports and react - neither of which we need. Humorously, they don't run the linter, and their files have indentation problems 🥇 . I think that says something for our strategy! Google just runs the linter as we do. They fail on warnings which is probably a good idea! Other than that, I feel like this is a lot like testing CSS. Sure, I could try to validate that all the things look correct—or not! 

The current strategy is actually just to validate that ESLint accepts and executes the configuration. The fact that the testing strategy is the linting strategy is an accidental, not-so-accidental approach. It will allow us test specific cases if we need to do so.

README.md
jattasNI: @alejandro5042 will do this

package.json
jattasNI: @rajsite will do this

README.md
jattasNI: This can wait until we've built up some best practices

packages/eslint-config/package.json
jattasNI: @rajsite will add the test to the pipeline. We don't think we need to add more tests.

README.md
```diff
@@ -0,0 +1,23 @@
+# NI JavaScript Style Guide
+
+Welcome to NI's internal and external JavaScript and TypeScript linter rules for [ESLint](https://eslint.org/docs/user-guide/getting-started) and [TSLint](https://palantir.github.io/tslint/).
```
rajsite, line 3: Could probably leave off saying `internal and external`

README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Installation
+
+```
+npm install --save-dev eslint eslint-plugin-import @ni/eslint-config
```
rajsite, line 8: Doesn't need to block this PR but should we consider the approach used by [airbnb](https://www.npmjs.com/package/eslint-config-airbnb) using a command like:
```
npx install-peerdeps --dev @ni/eslint-config
```

I believe the benefit is we get to align our rules with specific versions of eslint, etc.. With the current approach new eslint, airbnb, etc versions could break someone trying to setup a new repo.

README.md
```diff
@@ -0,0 +1,23 @@
...
+
+```json
+{
+    "extends": "@ni"
+}
+```
+
+## License
+
+[MIT (c) 2020 NI](./LICENSE)
```
rajsite, line 23: I think for legal stuff we are supposed to use [National Instruments Corporation](https://teams.microsoft.com/l/message/19:c0ff86b3253946e1948fe7089cd0bc91@thread.skype/1594757999233?tenantId=87ba1f9a-44cd-43a6-b008-6fdb45a5204e&groupId=35260d9f-7a49-4099-b8cc-13346dcbab4e&parentMessageId=1594743825112&teamName=Open%20Source&channelName=General&createdTime=1594757999233)

package.json
```diff
@@ -0,0 +1,24 @@
...
+  "description": "NI's internal and external JavaScript and TypeScript style guide",
```
rajsite, line 4: would leave off the internal and external here as well, unneeded qualifier

README.md
TrevorKarjanis: This was consistent with the Python style guide. However, I happen to agree, so I will remove it.

README.md
TrevorKarjanis: Hmm, npm install would install the latest of eslint and eslint-plugin-import, wouldn't it? That command does seem helpful, although, you probably already know the syntax if you're using it. I will follow Airbnb's suit and include both a mention of the peer dependencies and the shortcut.

As for enforcing versions, I had the same concern. It really depends on how the upstream package's semantic versioning is handled, and [ESLint's recommendation](https://eslint.org/docs/developer-guide/shareable-configs#publishing-a-shareable-config) is to use the ```>=``` operator. This is likely to allow the use of the latest version most commonly used with a new setup, but [they follow semantic versioning](https://eslint.org/docs/maintainer-guide/releases)! They will break us across major releases, so how about we use the minor release operator, e.g. ```^7.6.0```, for eslint and eslint-plugin-import and the patch release operator for eslint-config-airbnb-base, e.g. ```~14.2.0```?

rajsite: I like the cut of your jib; sounds like a good approach to me 👍

PR #2, opened 2020-09-15
Add resolved rules and comments.

TrevorKarjanis:
- Add rules and comments that have been resolved.

Notes
- The style of the rules definition in index.js will evolve as rules are added.

Linted and tested
- Cowa-friggin'-bunga

### DIFF
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -1,3 +1,10 @@
 module.exports = {
   extends: 'airbnb-base',
+  rules: {
+    /*
+        Errors from this rule will identify issues more often than not. However, there may be
+        unique exceptions where this rule can be disabled with an inline comment.
+    */
+    'no-await-in-loop': 'error',
+  },
 };
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -1,23 +1,36 @@
 module.exports = {
-  extends: 'airbnb-base',
-  rules: {
-    /*
-        This rule is controversial and is defined such that trailing commas are not required but
-        may be used if preferred.
-    */
-    'comma-dangle': ['error', 'only-multiline'],
-    /*
-        Errors from this rule will identify issues more often than not. However, there may be
-        unique exceptions where this rule can be disabled with an inline comment.
-    */
-    'no-await-in-loop': 'error',
-    /*
-        Console logs should not be exposed in production environments.
-    */
-    'no-console': 'error',
-    /*
-        Migrating from var is not trivial, so there may be exceptions for older projects.
-    */
-    'no-var': 'error',
-  },
+    extends: 'airbnb-base',
+    rules: {
+        /*
+            Omit arrow function parenthesis where they are not required to improve readability.
+        */
+        'arrow-parens': ['error', 'as-needed'],
+        /*
+            This rule is controversial and is defined such that trailing commas are not required but
+            may be used if preferred.
+        */
+        'comma-dangle': ['error', 'only-multiline'],
+        /*
+            Do not require a newline at the end of every file in favor of reducing the effort to
+            satisfy the rule over the proposed advantages on UNIX and with concatenation.
+        */
+        'eol-last': 'off',
+        /*
+            This indentation size is used to promote consistency.
+        */
+        indent: ['error', 4],
+        /*
+            Errors from this rule will identify issues more often than not. However, there may be
+            unique exceptions where this rule can be disabled with an inline comment.
+        */
+        'no-await-in-loop': 'error',
+        /*
+            Console logs should not be exposed in production environments.
+        */
+        'no-console': 'error',
+        /*
+            Migrating from var is not trivial, so there may be exceptions for older projects.
+        */
+        'no-var': 'error',
+    }
 };
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -2,6 +2,11 @@ module.exports = {
   extends: 'airbnb-base',
   rules: {
     /*
+        This rule is controversial and is defined such that trailing commas are not required but
+        may be used if preferred.
+    */
+    'comma-dangle': ['error', 'only-multiline'],
+    /*
         Errors from this rule will identify issues more often than not. However, there may be
         unique exceptions where this rule can be disabled with an inline comment.
     */
@@ -10,5 +15,9 @@ module.exports = {
         Console logs should not be exposed in production environments.
     */
     'no-console': 'error',
+    /*
+        Migrating from var is not trivial, so there may be exceptions for older projects.
+    */
+    'no-var': 'error',
   },
 };
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -20,6 +20,14 @@ module.exports = {
         */
         indent: ['error', 4],
         /*
+            This is the default value which is defined here to note that there may be exceptions
+            where it is appropriate to disable this rule for older projects or with a comment.
+            Including one class per file is a best practice and recommended by the Angular style
+            guide. However, migrating older projects may not be trivial, and there may be
+            exceptions for organization.
+        */
+        'max-classes-per-file': ['error', 1],
+        /*
             Errors from this rule will identify issues more often than not. However, there may be
             unique exceptions where this rule can be disabled with an inline comment.
         */
@@ -32,5 +40,10 @@ module.exports = {
             Migrating from var is not trivial, so there may be exceptions for older projects.
         */
         'no-var': 'error',
-    }
+        /*
+            This configuration already supports the JSDoc syntax. Add additional syntax as line or
+            block exceptions or markers when necessary.
+        */
+        'spaced-comment': 'error'
+    },
 };
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -6,5 +6,9 @@ module.exports = {
         unique exceptions where this rule can be disabled with an inline comment.
     */
     'no-await-in-loop': 'error',
+    /*
+        Console logs should not be exposed in production environments.
+    */
+    'no-console': 'error',
   },
 };
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -5,66 +5,90 @@ module.exports = {
             Omit arrow function parenthesis where they are not required to improve readability.
         */
         'arrow-parens': ['error', 'as-needed'],
+
         /*
             This rule is controversial and is defined such that trailing commas are not required but
             may be used if preferred.
         */
         'comma-dangle': ['error', 'only-multiline'],
+
         /*
             Do not require a newline at the end of every file in favor of reducing the effort to
             satisfy the rule over the proposed advantages on UNIX and with concatenation.
         */
         'eol-last': 'off',
+
         /*
             There may be exceptions where it is appropriate to disable this rule for older projects
             that make use of Immediately Invoked Function Expressions (IIFE).
         */
         'func-names': ['error', 'as-needed'],
+
         /*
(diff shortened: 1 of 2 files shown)


packages/eslint-config/index.js
```diff
@@ -1,3 +1,52 @@
 module.exports = {
-  extends: 'airbnb-base',
-};
+    extends: 'airbnb-base',
+    rules: {
+        'arrow-parens': ['error', 'as-needed'],
+        'comma-dangle': ['error', 'only-multiline'],
+        'eol-last': 'off',
+        /*
+            This is the default value which is defined here to note that there may be exceptions
```
jattasNI, line 8: I don't think this one matches the default for airbnb (which was `'warn'`) so we can delete the first sentence. I'd also like to add a bit more detailed rationale. How about: "Legacy applications that make heavy use of the IIFE (Immediately invoked function expression) may choose to disable this rule"

packages/eslint-config/index.js
```diff
@@ -1,3 +1,52 @@
...
+            where it is appropriate to disable this rule for older projects.
+        */
+        'func-names': ['error', 'as-needed'],
+        indent: ['error', 4],
+        'linebreak-style': 'off',
+        /*
+            This is the default value which is defined here to note that there may be exceptions
+            where it is appropriate to disable this rule for older projects or with a comment.
```
jattasNI, line 16: I propose changing "with a comment" to "in specific files" just to keep the focus on the "what" instead of the "how".

packages/eslint-config/index.js
```diff
@@ -1,3 +1,52 @@
...
+            Including one class per file is a best practice and recommended by the Angular style
+            guide. However, migrating older projects may not be trivial, and there may be
+            exceptions for organization.
```
jattasNI, line 19: I propose changing "there may be exceptions for organization" to something more specific. We could borrow the language from our C# conventions: "Exceptions may be made to the one "main" type per file rule for public/internal types that are only used as part of the interface to the main type and no other types."

packages/eslint-config/index.js
```diff
@@ -1,3 +1,52 @@
...
+        */
+        'no-prototype-builtins': 'error',
+        /*
+            This is the default value which is defined here to note that migrating from var is not
+            trivial, so there may be exceptions for older projects.
+        */
+        'no-var': 'error',
+        /*
+            This rule is controversial, but defined as is for consistency. Requiring a space before
+            function parenthesis improves searches in plain text environments.
```
jattasNI, line 47: Consider adding "... and is a common pattern in JavaScript codebases" just to convey that we're not making this up ourselves

I did some more reading on this and one of the more "authoritative" (if there is such a thing in the JS world) proponents of the space is Douglas Crockford of JavaScript the Good Parts.
https://www.crockford.com/code.html

jattasNI (changes requested):

jattasNI:
Have you already reached out to Carson/David/Tim to get their feedback on this PR even if they're not yet members of the /ni org?

packages/eslint-config/index.js
TrevorKarjanis: Thanks, that is helpful. Here is the [rationale for AirBnb's default value](https://github.com/airbnb/javascript/issues/55#issuecomment-101374665) which explains the inconsistency.

TrevorKarjanis: ... which unfortunately is aligned with Crockford's advice.
https://www.crockford.com/code.html#function

packages/eslint-config/index.js
TrevorKarjanis: That was my mistake, good catch.

packages/eslint-config/index.js
TrevorKarjanis: This is intentional. The distinction between projects and comments is to convey specifically whether the exception should be per instance (file or line) or generally. In this sense, I am including the "how" to provide additional information.

packages/eslint-config/index.js
TrevorKarjanis: Yes, your suggestion is better.

packages/eslint-config/index.js
jattasNI: I see. I'm probably nitpicking, but "with a comment" comes across a bit vague to me. Other suggestions: 
1. "with a file suppression comment" (I'm assuming this would always be disabled per-file...if not then "with a file or line suppression comment")
1. "in specific files with `// tslint:disable:max-classes-per-file`"

packages/eslint-config/index.js
rajsite: I think I'm at this point I'm willing to concede my vote on this idea. From the continued questions here and in the azdo thread it seems like what I believe are the actual tangible benefits (improved search) are not enough to settle an otherwise arbitrary decision.

If the current set of tangible benefits are not enough to stand on their own and extra justification is needed (well a lot of other people do it) then it's probably not worth doing. If we are not clearly convinced after all the discussion and a synchronous meeting and resolution then it will be difficult to convince the organization as a whole.

TrevorKarjanis:
> Have you already reached out to Carson/David/Tim to get their feedback on this PR even if they're not yet members of the /ni org?

The link is now available in the channel.

packages/eslint-config/index.js
TrevorKarjanis: I would be prefer to refer the reader to the ESLint documentation for the correct syntax, so I have gone with, "or by suppressing with a file or inline comment."

packages/eslint-config/index.js
TrevorKarjanis: @jattasNI Is that comment supposed to be "private/internal types"?

jattasNI: I think in the C# rules they're ok with even having multiple public types in a file if they are tightly coupled (an enum and the API that uses it or a factory and the type it produces, for example). 

If you think it's confusing, I would be ok with making it more vague and just removing "public/internal" in our case or adopting another suggestion if you have one.

packages/eslint-config/index.js
jattasNI: @rajsite our decision hasn't been sitting super well with me either now that we're trying to justify it. Are you changing your vote to 
1. airbnb default (aka Crockford's guideline)
1. something more aligned with C# (`space-before-function-paren: ["error", "never"]`)
1. something else

We can also bring this up at our next meeting if you want to revisit it with the whole group.

packages/eslint-config/index.js
```diff
@@ -1,3 +1,52 @@
 module.exports = {
-  extends: 'airbnb-base',
-};
+    extends: 'airbnb-base',
+    rules: {
+        'arrow-parens': ['error', 'as-needed'],
+        'comma-dangle': ['error', 'only-multiline'],
+        'eol-last': 'off',
+        /*
```
alejandro5042, line 7: I would prefer if we had new lines between all the rules, such that it looked like this (cleaner, better separation between rules):

```json
        'max-classes-per-file': ['error', 1],
        
        /*
            This is the default value which is defined here to note that there may be exceptions
            where it is appropriate to disable this rule with a comment. Awaits in loops are
            expected to be mistakes more often than not. However, there may be exceptions for
            unique situations.
        */
        'no-await-in-loop': 'error',
        
        'no-console': 'error',
        
        /*
            This is the default value which is defined here to note that there may be exceptions,
            like array iterators, where it is appropriate to disable this rule with a comment.
        */
        'no-loop-func': 'error',
        
        /*
            This is the default value which is defined here to note that this rule defends against
            prototype pollution vulnerabilities that exploitable with XSS.
        */
        'no-prototype-builtins': 'error',
```

packages/eslint-config/index.js
```diff
@@ -1,3 +1,52 @@
 module.exports = {
-  extends: 'airbnb-base',
-};
+    extends: 'airbnb-base',
+    rules: {
+        'arrow-parens': ['error', 'as-needed'],
```
alejandro5042, line 4: I would go as far as to say we should document each of our rules; aka. why did we decide to diverge.

packages/eslint-config/index.js
alejandro5042: I also find the original wording confusing and like Jesse's suggestion.

packages/eslint-config/index.js
```diff
@@ -1,3 +1,52 @@
...
+            This is the default value which is defined here to note that there may be exceptions
+            where it is appropriate to disable this rule for older projects.
+        */
+        'func-names': ['error', 'as-needed'],
+        indent: ['error', 4],
+        'linebreak-style': 'off',
+        /*
+            This is the default value which is defined here to note that there may be exceptions
+            where it is appropriate to disable this rule for older projects or with a comment.
+            Including one class per file is a best practice and recommended by the Angular style
```
alejandro5042, line 17: nitpick/minor:

> ...is a best practice and is also the recommendation by the Angular style guide...

Basically to separate the two a bit: it is an Angular style guide recommendation, but it is also, beyond that, a best practice in general.

packages/eslint-config/index.js
alejandro5042, line 8: > This is the default value which is defined here to note that...

I find this wording to be confusing. It's in most of the comments and I'm not sure what it means.

Wonder what value this phrase adds. Could we remove it?

```
/*
 It may be appropriate to disable this rule for older projects, such as when ...
*/

Comment (alejandro5042) [inline packages/eslint-config/index.js:8]: What are those exceptions?

Comment (alejandro5042) [reply packages/eslint-config/index.js:8]: OK I think this is the default value from Airbnb....


I wonder if we can reword as this:

```
/*
    This rule was copied from the Airbnb style guide without additional changes. We copy it here in order to add this comment:

    ...
*/
```

File: packages/eslint-config/index.js [hunk 8]
```diff
@@ -1,3 +1,52 @@
...
+            guide. However, migrating older projects may not be trivial, and there may be
+            exceptions for organization.
+        */
+        'max-classes-per-file': ['error', 1],
+        /*
+            This is the default value which is defined here to note that there may be exceptions
+            where it is appropriate to disable this rule with a comment. Awaits in loops are
+            expected to be mistakes more often than not. However, there may be exceptions for
+            unique situations.
```
alejandro5042, line 26: Those situations should be documented, right? _This_ file _is_ the NI style guide, which should give guidance. (Or we could link to a markdown file somewhere with additional examples?)

alejandro5042 (changes requested):
Overall, great! I would like to see the rationale brought out of our internal PR and into the open here, since this repo _is_ the style guide, and not my random internal PR which will be long forgotten next year.

packages/eslint-config/index.js
TrevorKarjanis: That is fine with me.

packages/eslint-config/index.js
jattasNI: We discussed this again in our meeting today and decided to go with no space. Here's my stab at the justification:

We believe that enforcing this rule enhances code readability by making code appearance more consistent across codebases. We therefore discourage codebases from disabling or changing the value of this rule.

We elect not to include a space before function parentheses. There are advocates in the external JavaScript community and internal NI codebases for both this configuration and configurations that include a space. Each of those opinions have valid merits. Our primary justification is that the no-space syntax will be more familiar to the many NI web developers with experience in other languages (C/C++/C# and Python, for example).

packages/eslint-config/index.js
TrevorKarjanis: I'm thinking I'll just remove the default value statements. They are repetitive and probably don't provide much value.

alejandro5042 (approved):

LukasKall (approved):

jattasNI (approved):

rajsite (approved):

PR #4, opened 2020-10-07
Add resolved rules and comments.

TrevorKarjanis:
- Add rules and comments that have been resolved.

Notes
- The style of the rules definition in index.js will evolve as rules are added.

Linted and tested
- Hakuna matata

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,6 +69,21 @@ module.exports = {
         'no-prototype-builtins': 'error',
 
         /*
+            Higher-order function composition, like Array.forEach(), is preferred for iterative
+            loops. For...in statements can be used when performance is a consideration as long as
+            guard-for-in is enabled. For...of statements can be used when break is required for
+            control flow or the iteration produces side-effects.
+        */
+        'no-restricted-syntax': [
+            'error', {
+                selector: 'LabeledStatement',
+                message: 'Labels are a form of GOTO; using them makes code confusing and hard to maintain and understand.',
+            }, {
+                selector: 'WithStatement',
+                message: '`with` is disallowed in strict mode because it makes code impossible to predict and optimize.',
+            }],
+
+        /*
             Migrating from var is not trivial, so there may be exceptions for older projects.
         */
         'no-var': 'error',
--- a/index.js
+++ b/index.js
@@ -146,12 +146,17 @@ module.exports = {
             Enforcing this rule improves code readability by making its appearance more consistent
             across codebases. Disabling or changing the value of this rule is discouraged.
 
-            We elect not to include a space before function parentheses. There are advocates in the
-            external JavaScript community and within NI for both this configuration and
-            others that include a space. Each of those opinions have valid merits. The primary
-            justification is that the syntax without a space will be more familiar to the many NI
-            web developers with experience in other languages (e.g. C, C++, C#, and Python).
+            We elect not to include a space before function parentheses except for anonymous
+            functions. There are advocates in the external JavaScript community and within NI for
+            both this configuration and others that include a space. Each of those opinions have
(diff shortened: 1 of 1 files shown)


rajsite:
Could we update the PR to reflect the in person discussion for `space-before-function-paren`:
```
'space-before-function-paren': ['error', { anonymous: 'always', named: 'never', asyncArrow: 'always' }]
```

which is reflected by the following [examples](https://eslint.org/docs/rules/space-before-function-paren#anonymous-always-named-never-asyncarrow-always):
```js
function foo() {
    // ...
}

var bar = function () {
    // ...
};

class Foo {
    constructor() {
        // ...
    }
}

var foo = {
    bar() {
        // ...
    }
};

var foo = async (a) => await a
```

TrevorKarjanis:
> Could we update the PR to reflect the in person discussion for `space-before-function-paren`:

@rajsite Yes, no problem.

rajsite:
@TrevorKarjanis looks like there are merge conflicts to resolve since the .eslintrc moved to the root

index.js
```diff
@@ -62,17 +74,68 @@ module.exports = {
...
+
         /*
             Migrating from var is not trivial, so there may be exceptions for older projects.
         */
         'no-var': 'error',
 
+        /*
+            Defining classes and variables before their use can cause errors. However, defining
+            functions and methods before their use is a common programming practice for
+            readability.
```
rajsite, line 124: "defining functions" is a little vague, ie a variable binding to a function expression is defining a function and would not be allowed. Also the position as written seems backwards from the decision for functions.

Do you think the following may be clearer?: "However, placing function declarations at the end of a file is common programming practice for readability"

rajsite (approved):

index.js
```diff
@@ -62,17 +74,68 @@ module.exports = {
...
         /*
             This rule defends against prototype pollution vulnerabilities that exploitable with
             XSS.
         */
         'no-prototype-builtins': 'error',
 
+        /*
+            Higher-order function composition is preferred for iterative loops. For...in statements
+            can be used when performance is a consideration as long as guard-for-in is enabled.
+            For..of statements can be used when break is required for control flow or the iteration
```
jattasNI, line 104: My nitpicking mind can't stop seeing that we use three dots in `for...in` but two in `for..of`. Can we use three for both?

index.js
```diff
@@ -62,17 +74,68 @@ module.exports = {
...
+        }],
+
         /*
             This rule defends against prototype pollution vulnerabilities that exploitable with
             XSS.
         */
         'no-prototype-builtins': 'error',
 
+        /*
+            Higher-order function composition is preferred for iterative loops. For...in statements
```
jattasNI, line 102: The phrase "higher-order function composition" might not be immediately understood for some readers (cough...me...cough) so I think adding an example would help. Something like
"Higher-order function composition (array.forEach(...)) is preferred for iterative loops".

Or we could even provide explicit guidance by stealing the guidance from airbnb's readme.
    > Use `map()` / `every()` / `filter()` / `find()` / `findIndex()` / `reduce()` / `some()` / ... to iterate over arrays, and `Object.keys()` / `Object.values()` / `Object.entries()` to produce arrays so you can iterate over objects.

jattasNI (approved):

LukasKall (approved):

mure (approved):

rajsite:
@alejandro5042 @dschwartzni @trstamper can y'll take a look when you have a chance?
P.S. I've submitted a request to have David and Tim added as collaborators in a public repo

alejandro5042 (approved):

index.js
TrevorKarjanis: Whoops! Yes, that is fine.

PR #10, opened 2020-11-20
Add no-param-reassign config

rajsite:
Updates the `no-param-reassign` configuration to treat function parameters as `const` bindings.

Testing locally, I was not able to just set `'no-param-reassign': 'error'` as it appears that the exisiting eslint config which sets `'no-param-reassign': ['error', {props: true, ...}]` was overriding the default configuration.

To allow assignment to properties of parameters (ie `const` behavior) I explicitly set `props: false`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -101,8 +101,16 @@ module.exports = {
         }],
 
         /**
-         * Prefer the '+=' operator. Allow unary operators in for loops, because it is a common
-         * pattern.
+            This rule disallows assigning to function parameters; function parameters are
+            treated as const bindings. Some more readable or type-safe alternatives:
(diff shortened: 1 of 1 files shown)


rajsite:
Also notifying folks not in org yet to comment on review @dschwartzni @trstamper

TrevorKarjanis (approved):

alejandro5042 (approved):

LukasKall (approved):

jattasNI (approved):
Thanks for double checking the `props` case and fixing the bug with our initial proposal.

mure (approved):

PR #14, opened 2021-01-04
Add recommended TypeScript rules part 1

mure:
This PR adds the set of recommended rules from the typescript-eslint package that do NOT require type checking. Pulled from here: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended.ts

I removed the rules that conflicted with the ones we already had in the file. 
You can find the table of all rules here to understand their usage:
https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#supported-rules

### DIFF
--- a/typescript.js
+++ b/typescript.js
@@ -1,5 +1,5 @@
 module.exports = {
-    extends: './index',
+    extends: ['./index', 'plugin:@typescript-eslint/recommended'],
     parser: '@typescript-eslint/parser',
     rules: {
         /*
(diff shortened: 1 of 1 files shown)


typescript.js
```diff
@@ -63,6 +63,27 @@ module.exports = {
             asyncArrow: 'always'
         }],
         'space-infix-ops': 'off',
-        '@typescript-eslint/space-infix-ops': 'error'
+        '@typescript-eslint/space-infix-ops': 'error',
+        /*
```
TrevorKarjanis, line 67: These rules are more likely of interest to readers than the extensions rules. Therefore, they are higher priority, and I think it would be intuitive and an advantage to place them first.

TrevorKarjanis (commented):
The extension rules were included directly, because I did not find a configuration to extend. However, why don't we just extend ```plugin:@typescript-eslint/recommended```? We are also able to extend ```eslint-recommended``` and ```recommended-requiring-type-checking``` as needed.

mure:
@TrevorKarjanis I made the config extend from `plugin:@typescript-eslint/recommended` and removed the hardcoded rules. The one side effect is that this PR will be a little harder to review now, because there's no rules to leave comments on. If we decide to make any overrides, I'll add them at the top of the config and put the comment back in.

TrevorKarjanis:
@mure Hmm, I had not considered that for the purposes of evaluating the rules here, but I'm not in favor of including the rules directly just to satisfy that purpose. I agree if we have changes now we can add them to this review or open issues afterward. If desired, we can include a comment with the link to the configuration files on GitHub. Milan already requested a similar comment for the extension rules.

jattasNI (approved):

jattasNI:
I added other reviewers as an FYI. I don't think we need everyone's approval in order to merge this since we discussed it in our meeting today.

rajsite (approved):

TrevorKarjanis (approved):

LukasKall (approved):

PR #16, opened 2021-01-13
Use  @typescript-eslint/all for now to help evaluate all lint rules

trstamper:
This PR changes "@typescript-eslint/recommended" to "@typescript-eslint/all" so that we can start with a very strict ruleset and whittle down to the rules we actually want.

### Testing: 
I tested that I could use this change in an existing typescript codebase and see errors for rules that aren't included in the recommended set (e.g. @typescript-eslint/explicit-function-return-type )

### DIFF
--- a/typescript.js
+++ b/typescript.js
@@ -1,5 +1,5 @@
 module.exports = {
-    extends: ['./index', 'plugin:@typescript-eslint/recommended'],
+    extends: ['./index', 'plugin:@typescript-eslint/all'],
     parser: '@typescript-eslint/parser',
     rules: {
         /*


TrevorKarjanis (commented):
I actually don't think this is a good idea. While enabling all the rules helps in evaluating the ruleset, that value depletes completely after the fact. Then the configuration becomes significantly less maintainable as upgrades are subject to all new rules in addition to changes to defaults. I also don't think this is so much more strict than it is just an opinionated set of defaults. It complicates our ability to separate the rules that require type information. Lastly, given its previous mentioned issues and that we can simply extend ```plugin:@typescript-eslint/all``` ourselves, it doesn't really simplify our ability to evaluate all the rules.

PR #17, opened 2021-01-14
Add TypeScript support for eslint-plugin-import.

TrevorKarjanis:
- Add TypeScript support for eslint-plugin-import to the TypeScript configuration.
- Disable import rules that are already handled by the TypeScript compiler.

Linted and tested
- Verified that TypeScript imports are linted correctly

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,8 +31,7 @@
   ],
   "dependencies": {
     "@typescript-eslint/parser": "^4.9.1",
-    "eslint-config-airbnb-base": "~14.2.0",
-    "eslint-import-resolver-typescript": "^2.3.0"
+    "eslint-config-airbnb-base": "~14.2.0"
   },
   "devDependencies": {
     "eslint": "^7.6.0",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
   ],
   "dependencies": {
     "@typescript-eslint/parser": "^4.9.1",
-    "eslint-config-airbnb-base": "~14.2.0"
+    "eslint-config-airbnb-base": "~14.2.0",
+    "eslint-import-resolver-typescript": "^2.3.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -31,7 +31,8 @@
   ],
   "dependencies": {
     "@typescript-eslint/parser": "^4.9.1",
-    "eslint-config-airbnb-base": "~14.2.0"
+    "eslint-config-airbnb-base": "~14.2.0",
+    "eslint-import-resolver-typescript": "^2.3.0"
```
rajsite, line 35: Should this be a peer dependency as it modifies the behavior of `eslint-plugin-import` which is also a peer dependency?

Probably `@typescript-eslint/parser` should be a peer dependency as well (it is a plugin for eslint)

TrevorKarjanis: Actually, it depends on how module is loaded. "If your shareable config depends on a third-party parser or another shareable config, you can specify these packages as dependencies." In the process of evaluating the TypeScript resolver I discovered that eslint-plugin-import actually includes a configuration that makes eslint-import-resolver-typescript unnecessary. I have tested it and updated the PR.

rajsite: Ah I see the [full context](https://eslint.org/docs/developer-guide/shareable-configs#publishing-a-shareable-config) was:

> If your shareable config depends on a plugin, you should also specify it as a peerDependency (plugins will be loaded relative to the end user's project, so the end user is required to install the plugins they need). However, if your shareable config depends on a third-party parser or another shareable config, you can specify these packages as dependencies.

Thanks!

rajsite: Was wondering how `import/typescript` worked and the [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import#typescript) docs clarified. Looks good to me!

rajsite (approved):

PR #18, opened 2021-01-20
Enable TypeScript Rules (without requiring type checking rules)

rajsite:
The goal of this PR is to be used for evaluating the TypeScript Rules used in `typescript.js`.

To test locally, you can add the devDependencies to a project with:
```
npm install -D @typescript-eslint/eslint-plugin@4.9.1 eslint@7.6.0 eslint-plugin-import@2.22.0
```
And extract the tgz from the npm package build artifact (see below) and run:
```
npm install -D ni-eslint-config-1.0.0-alpha.11.tgz
```

This PR also makes a few additional changes:
- Adds a tool usable via `npm run dev-print-typescript-props` to print the properties expected in each of the TypeScript configurations based on the installed typescript-eslint version. This helps in maintenance of tracking changes to the typescript-extension rules.
- Adds basic smoke tests for the typescript and typescript-requiring-type-checking configurations
- Adds the typescript-requiring-type-checking rules (the rules themselves are not target for discussion in this PR and are currently commented out).
- TypeScript Extension rules were separated into different files and split by those rules requiring type checking and those that don't. In addition they were synced with the current NI configuration / Airbnb configuration that defines them for eslint.

Build artifacts:
- [Initial PR](https://github.com/ni/javascript-styleguide/actions/runs/499422317)
- [First update with resolved rules (3/10/21)](https://github.com/ni/javascript-styleguide/actions/runs/641593158)
- [Second update with resolved rules (3/11/21)](https://github.com/ni/javascript-styleguide/actions/runs/643789555)


typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
-        }],
-        'space-infix-ops': 'off',
-        '@typescript-eslint/space-infix-ops': 'error',
+
+        // None
+
+        /*
+            TypeScript rules outside of the recommended configuration
+        */
+        '@typescript-eslint/array-type': 'error',
```
rajsite, line 27: Test

rajsite: proposal: be done

jattasNI (changes requested):

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/class-literal-property-style': 'error',
+        '@typescript-eslint/consistent-indexed-object-style': 'error',
+        '@typescript-eslint/consistent-type-assertions': 'error',
+        '@typescript-eslint/consistent-type-definitions': 'error',
+        '@typescript-eslint/consistent-type-imports': 'error',
+        '@typescript-eslint/explicit-function-return-type': 'error',
+        '@typescript-eslint/explicit-member-accessibility': 'error',
+        '@typescript-eslint/member-delimiter-style': 'error',
+        '@typescript-eslint/member-ordering': 'error',
+        '@typescript-eslint/method-signature-style': 'error',
```
LukasKall, line 38: I tend to disable [@typescript-eslint/method-signature-style](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/method-signature-style.md). I like to use the same syntax in interface definitions as in class definitions, which is also more common in other languages used by NI like C#.
```typescript
interface T1 {
    func(arg: string): number;
}

class C1 implements T1 {
    func(arg: string): number {
    }
}
```
Using the property style looks a bit like defining a property on an interface, which is not the case.
```typescript
interface T2 {
    func: (arg: string) => number;
}
```
Function overloads also look kind of confusing.
```typescript
interface T3 {
  func: ((arg: number) => void) &
    ((arg: string) => void) &
    ((arg: boolean) => void);
}
```
I wouldn't use the `"method"` option for the rule, because in some cases you probably want to use the property style to make use of the TypeScript option `strictFunctionTypes`.

lib/typescript-extensions.js
```diff
@@ -0,0 +1,169 @@
...
+        'no-redeclare': 'off',
+        '@typescript-eslint/no-redeclare': 'error',
+
+        // Defined by Airbnb
+        'no-shadow': 'off',
+        '@typescript-eslint/no-shadow': 'error',
+
+        // Defined by Airbnb
+        'no-unused-expressions': 'off',
+        '@typescript-eslint/no-unused-expressions': ['error', {
```
jattasNI, line 126: **@typescript-eslint/no-unused-expressions**
[This rule](https://eslint.org/docs/rules/no-unused-expressions) flagged some Web Portal code which looked fine to me:
```ts
    ngOnDestroy(): void {
        this.selectionChangedSubscription && this.selectionChangedSubscription.unsubscribe();
   }
```
It's counting the method body as an "unused expression" but I like the short circuit function call syntax.

**Proposal**: set `allowShortCircuit` and `allowTernary` to `true.

jattasNI (changes requested):

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/no-confusing-non-null-assertion': 'error',
+        '@typescript-eslint/no-dynamic-delete': 'error',
+        '@typescript-eslint/no-extraneous-class': 'error',
+        '@typescript-eslint/no-implicit-any-catch': 'error',
+        '@typescript-eslint/no-invalid-void-type': 'error',
+        '@typescript-eslint/no-parameter-properties': 'error',
```
LukasKall, line 44: **[@typescript-eslint/no-parameter-properties](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-parameter-properties.md)**
Parameter properties are very common in Angular and I tend to use them in non Angular projects as well. I see the rationale that they are less explicit and clear than other declarations, but at least for `private readonly` parameters I feel they are pretty usefull.

**Proposal**

```javascript
/*
    Parameter properties are less explicit than other ways of declaring and initializing class members,
    but it's a nice shorthand for properties ingested in the constructor, which are only accessible
    in their containing class.
*/
'@typescript-eslint/no-parameter-properties':  ['error', { 'allows': ['private readonly'] }]
```

jattasNI (changes requested):

typescript.js
jattasNI: I agree, I had the same reaction when trying out this rule with the `"property"` option.

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/no-require-imports': 'error',
+        '@typescript-eslint/no-type-alias': 'error',
```
jattasNI, line 46: [@typescript-eslint/no-type-alias](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-type-alias.md)
The Web Portal codebase has a pattern of using type aliases in unions to form string enums:
```ts
export type HighlightRuleType = 'lt' | 'gt' | 'between';
```
This seems fine to me and we could allow it via `"allowAliases": "in-unions"`

I also saw a handful of uses to make code more readable, especially in ngrx reducers with generics:
```ts
export type DataSourceSchemaState = EntityState<DataSourceSchema>;
```
After reading the docs I don't understand what the problem is with the above code so I'm tempted to disable the rule entirely, but maybe I'm missing something.

jattasNI (changes requested):

lib/typescript-extensions.js
```diff
@@ -0,0 +1,169 @@
...
+
+        // Defined by NI
+        'no-use-before-define': 'off',
+        '@typescript-eslint/no-use-before-define': ['error', { functions: false, classes: true, variables: true }],
+
+        // Defined by Airbnb
+        'no-useless-constructor': 'off',
+        '@typescript-eslint/no-useless-constructor': 'error',
+
+        // Available in newer typescript-eslint version
```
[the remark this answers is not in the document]
rajsite: Talked offline but I had a newer version of tools installed locally than what is currently specified in the package.json.

We could either disable it now for TypeScript and in a follow-on PR enable it along with version updates for related tools (eslint had a major version change) or I can do the updates as part of this PR.

I'd lean towards a follow-up PR as to not add noise or confusion here but don't feel strongly.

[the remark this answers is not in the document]
rajsite: Proposal: `'object-curly-spacing': 'off'` for this PR.

And in a follow-up PR with updated versions keep `'object-curly-spacing': 'off'` but enable `@typescript-eslint/object-curly-spacing`.

typescript.js
jattasNI: Oops, I left a comment agreeing with you but accidentally deleted it. To summarize:

In fact I think we should discuss going further turning off this rule entirely. I haven't seen any confusion in classes that define public parameter properties. It's a new syntax to learn and you have to remember to look in 2 places if you want to find all the properties, but for me those aren't strong enough reasons to ban the syntax.

**Proposal**
`'@typescript-eslint/no-parameter-properties':  'off' // Parameter properties are a nice shorthand for defining properties ingested in the constructor`

lib/typescript-extensions.js
rajsite: I think this would be better written using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) as:
```js
this.selectionChangedSubscription?.unsubscribe();
```

There was also [previous discussion on this property in the js rules](https://ni.visualstudio.com/Users/_git/abarreto-misc/pullrequest/83582?_a=files&path=%2Fairbnb-js-review%2Frules%2Fbest-practices.js) where we decided to keep this configuration off.

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        */
+        '@typescript-eslint/array-type': 'error',
+        '@typescript-eslint/ban-tslint-comment': 'error',
+        '@typescript-eslint/class-literal-property-style': 'error',
+        '@typescript-eslint/consistent-indexed-object-style': 'error',
+        '@typescript-eslint/consistent-type-assertions': 'error',
+        '@typescript-eslint/consistent-type-definitions': 'error',
+        '@typescript-eslint/consistent-type-imports': 'error',
+        '@typescript-eslint/explicit-function-return-type': 'error',
+        '@typescript-eslint/explicit-member-accessibility': 'error',
```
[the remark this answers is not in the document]
mure: I believe requiring accessibility modifiers would be in line with our C# rules.

mure:
If anyone else is doing this on a Mac, note that the built-in Archive Utility seems to try to recursively unpackage files. So I was very confused as to where `ni-eslint-config-1.0.0-alpha.11.tgz` was, but it turns out the utility was automatically unpacking that as well.

mure:
[@typescript-eslint/no-invalid-void-type](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-invalid-void-type.md)

I'm still wading my way through the Document Manger lint errors, but one of the interesting examples that came up was this piece of an observable chain:
```
switchMap((webapp: WebApp): Observable<[Document, void]> => combineLatest([this.getDocumentFromWebApp(webapp), this.loadDocuments(this.activePlugin)])),
```

On a glance, it seems reasonable - it's a `combineLatest` where one of the observables is void. However, I'm not sure if there's a better method in RxJS for accomplishing the same thing while avoiding the void type. We could consider turning on `allowInGenericTypeArguments` for these use cases.

typescript.js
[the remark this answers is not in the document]
LukasKall: I would leave this rule on. It let's you think, whether a class member is public or private which causes cleaner public APIs. If you're lazy and write every member without accessibility modifier, every member is public which may not be intended.

LukasKall:
> [@typescript-eslint/no-invalid-void-type](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-invalid-void-type.md)
> 
> I'm still wading my way through the Document Manger lint errors, but one of the interesting examples that came up was this piece of an observable chain:
> 
> ```
> switchMap((webapp: WebApp): Observable<[Document, void]> => combineLatest([this.getDocumentFromWebApp(webapp), this.loadDocuments(this.activePlugin)])),
> ```
> 
> On a glance, it seems reasonable - it's a `combineLatest` where one of the observables is void. However, I'm not sure if there's a better method in RxJS for accomplishing the same thing while avoiding the void type. We could consider turning on `allowInGenericTypeArguments` for these use cases.

I'm not sure, but is this something where `never` can be used? Or does this cause an error because `loadDocuments` returns `void`?

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/member-delimiter-style': 'error',
+        '@typescript-eslint/member-ordering': 'error',
+        '@typescript-eslint/method-signature-style': 'error',
+        '@typescript-eslint/no-confusing-non-null-assertion': 'error',
+        '@typescript-eslint/no-dynamic-delete': 'error',
+        '@typescript-eslint/no-extraneous-class': 'error',
+        '@typescript-eslint/no-implicit-any-catch': 'error',
```
LukasKall, line 42: [@typescript-eslint/no-implicit-any-catch](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-implicit-any-catch.md)
Typescript doesn't support catching specific errors. With 4.0 it added support for catching `:any` or `:unknown`, but I don't see much benefit in catching `:unknown`, because you would have to use [Type Guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) or something similar to use the catched error. I suggest to disable this rule. We already enable the rule [no-throw-literal](https://eslint.org/docs/rules/no-throw-literal), which forces to only throw `Error` objects. So you can be sure that the catched object is at least an instance of `Error`, which is totally sufficient most of the time.

**Proposal**
`'@typescript-eslint/no-implicit-any-catch': 'off'`

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/no-invalid-void-type': 'error',
+        '@typescript-eslint/no-parameter-properties': 'error',
+        '@typescript-eslint/no-require-imports': 'error',
+        '@typescript-eslint/no-type-alias': 'error',
+        '@typescript-eslint/no-unnecessary-type-constraint': 'error',
+        '@typescript-eslint/no-unused-vars-experimental': 'error',
+        '@typescript-eslint/prefer-enum-initializers': 'error',
```
mure, line 49: [@typescript-eslint/prefer-enum-initializers](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-enum-initializers.md)

Personally, I'm not a fan of this rule. I think it's cleaner if enum values are invisible, and I don't want to have to come up with a value for all of them. 

The justification is:
>However, by implicitly defining values, enums can lead to unexpected bugs if it's modified without paying attention to the order of its items.

I can't imagine a scenario where this would matter unless you are directly referencing an enum's value, which would be pretty bad code design. Does anyone have a real-world example to change my mind?

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        /*
+            TypeScript rules outside of the recommended configuration
+        */
+        '@typescript-eslint/array-type': 'error',
+        '@typescript-eslint/ban-tslint-comment': 'error',
+        '@typescript-eslint/class-literal-property-style': 'error',
+        '@typescript-eslint/consistent-indexed-object-style': 'error',
+        '@typescript-eslint/consistent-type-assertions': 'error',
+        '@typescript-eslint/consistent-type-definitions': 'error',
+        '@typescript-eslint/consistent-type-imports': 'error',
```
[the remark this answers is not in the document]
jattasNI: Consensus among Milan/Carson/Lukas/Jesse is that this repo should stay up to date with the latest TypeScript version so it's not a barrier to codebases who want to stay up to date.

It might also be helpful to add documentation that describes which rules target newer features to help repos find and disable. As a starting point, lets annotate rules that we notice requiring new TS features (like this one) in the README.md.

lib/typescript-extensions.js
```diff
@@ -0,0 +1,169 @@
...
+            overrides: {
+                return: { after: true },
+                throw: { after: true },
+                case: { after: true }
+            }
+        }],
+
+        // Defined by Airbnb
+        'lines-between-class-members': 'off',
+        '@typescript-eslint/lines-between-class-members': ['error', 'always', { exceptAfterSingleLine: false }],
```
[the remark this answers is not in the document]
jattasNI: We were seeing this rule applied on more than just functions, but we're confused about whether that's intended behavior. We should see if that's a bug or not by comparing the behavior to the JS version of the rule. I'll investigate further.

typescript.js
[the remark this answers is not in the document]
TrevorKarjanis: Support for this rule was [only recently added](https://github.com/typescript-eslint/typescript-eslint/issues/2559) for DI languages. I validated it linting Angular with @typescript-eslint/eslint-plugin 4.15.0, the latest version.

lib/typescript-extensions.js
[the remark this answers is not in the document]
jattasNI: The ESLint version of the rule only officially supports functions, not fields, because class fields aren't a "stage 4" JavaScript feature yet (see comment in [this issue](https://github.com/eslint/eslint/issues/13295)). That said, their implementation works for fields.

The typescript-eslint version of the rule doesn't explicitly add capabilities for fields in TypeScript; it just inherits the capability of the ESLint rule.

So we're in a weird limbo where the rule works for fields but neither package wants to take credit for it so I'm not sure where the docs stating this support should belong. I'm not planning to submit an issue or PR to either repo unless someone feels strongly that I should. The proposal above will continue to work well for our use case.

typescript.js
jattasNI: I've occasionally seen enums being cast to ints in C# code so I could imagine the same happening in TypeScript code, but I agree it's not the safest thing to do. One possible use case might be for serializing an enum's value to JSON.

But neither of those are particularly compelling to me so I'm ok to disable the rule unless we find a compelling example.

typescript.js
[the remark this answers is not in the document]
TrevorKarjanis: Type imports are for edge cases and, ["[are] something most users may never have to think about"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html). The two cases where they are used are [modules with side-effects](https://github.com/microsoft/TypeScript/issues/9191) and [file by file transpiling](https://github.com/microsoft/TypeScript/pull/35200). These are not the common cases, and their use and this rule should be used on a case by case basis. Otherwise, declaring type imports are not helpful. In fact, it's incredibly annoying to have to correlate and update the form of import with a type's usage which can change. Additionally, it's commonly unclear which type is actually elided when imports include used and unused types from the same file.

**Proposal**

Leave this rule unconfigured.

lib/typescript-extensions.js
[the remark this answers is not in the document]
TrevorKarjanis: Enforcing lines between methods is common, because methods tend to be multiline, complex, and yet filled with whitespace. Lines between methods enforce boundaries and consistency, and I place lines between single line methods as well for the latter reason. However, lines between fields is not as common, because they are generally equivalent to declaring variables. Fields tend to be a single line the extra whitespace superfluous, even noisy. It also prevents grouping.

This is probably why the TS rule doesn't address fields. The common desire is to enforce space between methods. After testing, exceptAfterSingleLine doesn't even enforce multiline fields.

typescript.js
TrevorKarjanis: Additionally, it would be contradictory to allow private properties but disallow public.
```typescript
constructor(private service: Service, bps: BreakpointService) {
  this.bps = bps;
}
```

typescript.js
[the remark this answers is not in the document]
TrevorKarjanis: In practice, implied public properties have been plenty clean, particularly with additional rules that require sorting. In fact, the lack of a modifier differentiates them more. Lazy consideration for accessibility is a training problem that may warrant failure. However, I could see an argument for private unless explicitly public.

I don't think this rule adds significant value, but I will remain open to the consensus.

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/explicit-function-return-type': 'error',
+        '@typescript-eslint/explicit-member-accessibility': 'error',
+        '@typescript-eslint/member-delimiter-style': 'error',
+        '@typescript-eslint/member-ordering': 'error',
+        '@typescript-eslint/method-signature-style': 'error',
+        '@typescript-eslint/no-confusing-non-null-assertion': 'error',
+        '@typescript-eslint/no-dynamic-delete': 'error',
+        '@typescript-eslint/no-extraneous-class': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: There are a few problems with using objects in place of static classes. Developers that use static classes don't think about them like objects. They expect to use accessibility modifiers and method shorthand (which contradicts our [guidance for objects](https://github.com/ni/javascript-styleguide/blob/c6bd2da9ff6d08cafc86ee12ef5f7bf5a20263d5/index.js#L175)). They also expect to use colon type syntax, equal operator assignment, and uninitialized fields.
```typescript
private instance: Class;
public field = '';
method(): void { }
```
I would think we would want ```allowStaticOnly``` or a decision to consolidate on [namespaces](https://www.typescriptlang.org/docs/handbook/namespaces.html#namespacing).

typescript.js
TrevorKarjanis: Types are absolutely useful, and I didn't find their examples convincing. Claiming the feature should be disabled, because there are bad practices isn't a good argument. Examples like primitive unions are common type definitions, so why not allow an alias? Many of the union and intersection examples don't reflect real world cases, e.g ```string & string[]```. Do we really have someone that would write  ```type myString = string```? Alias unions, callbacks, mapped types, and tuples are common. I've actually used constructors for mixins. I do agree that interfaces are preferred over object literals.

**Proposal**
Disallow object literals in favor of interfaces except where explicitly disabled with a comment.

```typescript
/* 
  Type aliases should be used to promote readable and DRY coding practices. However, interfaces should be preferred
  over object literals.
*/
'@typescript-eslint/no-type-alias': ['error', {
  allowAliases: 'always',
  allowCallbacks: 'always',
  allowConditionalTypes: 'always',
  allowConstructors: 'always',
  allowLiterals: 'never',
  allowMappedTypes: 'always',
  allowTupleTypes: 'always'
}]
```

TrevorKarjanis:
> I'm not sure, but is this something where `never` can be used? Or does this cause an error because `loadDocuments` returns `void`?

Never implies that a value is ```never``` provided, like if the observable always errored in which case no values would ever be provided. I'll have to run this rule over my code, but assigning ```void``` to variables [doesn't make much sense](https://www.typescriptlang.org/docs/handbook/basic-types.html#void). In @mure 's case, ```combineLatest``` expects values, so unless it is mapped to a single value then ```undefined``` is probably correct.

typescript.js
TrevorKarjanis: Shouldn't we be utilizing the type checker by defining the type, i.e. ```catch (e: Error)```? Are you more in favor of allowing implicit ```any```?

typescript.js
TrevorKarjanis: Yes, the real world use case is when an enum crosses a code boundary. The common example being values passed between client and server. In these cases the values need to be preserved. However, it's more common for enums to be used only in the client where numerical values are inferred and cleanup is avoided. I use implicit enums all the time and prefer the option as well.

typescript.js
LukasKall: I think that is not supported by TypeScript, otherwise I would be in favor of specifying the type to catch. 

image

TrevorKarjanis: Thanks Lukas, you are right ([#20024](https://github.com/microsoft/TypeScript/issues/20024)).

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/no-implicit-any-catch': 'error',
+        '@typescript-eslint/no-invalid-void-type': 'error',
```
mure, line 43: [@typescript-eslint/no-invalid-void-type](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-invalid-void-type.md)

I'm still wading my way through the Document Manger lint errors, but one of the interesting examples that came up was this piece of an observable chain:
```
switchMap((webapp: WebApp): Observable<[Document, void]> => combineLatest([this.getDocumentFromWebApp(webapp), this.loadDocuments(this.activePlugin)])),
```

On a glance, it seems reasonable - it's a `combineLatest` where one of the observables is void. However, I'm not sure if there's a better method in RxJS for accomplishing the same thing while avoiding the void type. We could consider turning on `allowInGenericTypeArguments` for these use cases.

mure: @LukasKall said:
I'm not sure, but is this something where never can be used? Or does this cause an error because loadDocuments returns void?

mure: @TrevorKarjanis said:
Never implies that a value is never provided, like if the observable always errored in which case no values would ever be provided. I'll have to run this rule over my code, but assigning void to variables doesn't make much sense. In @mure 's case, combineLatest expects values, so unless it is mapped to a single value then undefined is probably correct.

mure: I don't think using undefined is possible in this case, since it has to be a type argument. I get a compiler error:
`Type 'void' is not assignable to type 'undefined'`

jattasNI: @mure will continue researching this and come up with a proposal.

typescript.js
jattasNI: @mure to add a proposal to disable this rule

typescript.js
jattasNI: @rajsite this thread is awaiting your vote

typescript.js
jattasNI: I'll double check that Trevor's proposed configuration would allow the cases I described in my original comment but otherwise there doesn't seem to be any objection to Trevor's proposal.

typescript.js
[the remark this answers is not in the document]
jattasNI: @TrevorKarjanis to create a proposal comment including both this static case and the decorator case.

typescript.js
jattasNI: @rajsite 🚀

lib/typescript-extensions.js
jattasNI: @rajsite 🚀

lib/typescript-extensions.js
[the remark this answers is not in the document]
jattasNI: @rajsite 🚀
We also need to make the same change to the eslint config.

typescript.js
jattasNI: @rajsite 🚀

typescript.js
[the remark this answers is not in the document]
jattasNI: @rajsite 🚀

lib/typescript-extensions.js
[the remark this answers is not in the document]
jattasNI: @rajsite 🚀

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/ban-tslint-comment': 'error',
+        '@typescript-eslint/class-literal-property-style': 'error',
+        '@typescript-eslint/consistent-indexed-object-style': 'error',
+        '@typescript-eslint/consistent-type-assertions': 'error',
+        '@typescript-eslint/consistent-type-definitions': 'error',
+        '@typescript-eslint/consistent-type-imports': 'error',
+        '@typescript-eslint/explicit-function-return-type': 'error',
+        '@typescript-eslint/explicit-member-accessibility': 'error',
+        '@typescript-eslint/member-delimiter-style': 'error',
+        '@typescript-eslint/member-ordering': 'error',
```
TrevorKarjanis, line 37: [@typescript-eslint/member-ordering](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-ordering.md)
This rule groups decorated and abstract members together. This is problematic if not incorrect. If decorated members are grouped it's always because of their type (e.g. Inputs and Outputs). That is not always the case (e.g. HostListener), because decorators are modifiers that don't always imply a relationship with each other or a stronger relationship over other attributes. At the minimum I would group decorated members with their accessibility group members. It might seem reasonable to group public decorated fields together in Angular, because they tend to be Inputs and Outputs. However, that may not make sense with any other decorator or language semantics other than Angular. Therefore, I think it's best not to even group decorated members within accessibility groups, and leave it up to the developer's understanding of their semantics.

As for abstract members, we organize them by their relationship to members within the API as if they aren't abstract and as they would be ordered in documentation. I think they should be handled similarly to decorated members.

This proposal better aligns with our [C# guidelines](https://github.com/ni/csharp-styleguide#c6-%EF%B8%8F-consider-arranging-types-and-members-in-logical-layers) which take an even less restrictive approach in favor of using good judgement to achieve readability. It still organizes by accessibility as the [default](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-ordering.md#default-configuration) and the current TSLint rule do. Unfortunately, accessors are [not yet supported](https://github.com/typescript-eslint/typescript-eslint/issues/929), so I included a recommendation for them in the comment.

```typescript
/*
  Group members by fields and methods and then order them by accessibility starting with statics. Order members
  within these groups in a logical organization where readability is the most important thing. Private fields
  that back public accessors may be grouped with their accessor by disabling the rule with a comment.
*/
'@typescript-eslint/member-ordering': ['error', {
  default: [
    'signature',
    'public-static-field',
    'protected-static-field',
    'private-static-field',
    'static-field',
    'public-field',
    'protected-field',
    'private-field',
    'field',
    'public-constructor',
    'protected-constructor',
    'private-constructor',
    'constructor',
    'public-static-method',
    'protected-static-method',
    'private-static-method',
    'static-method',
    'public-method',
    'protected-method',
    'private-method',
    'method'
  ]
}]
```

TrevorKarjanis (changes requested):

typescript.js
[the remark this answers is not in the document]
TrevorKarjanis: Coincidentally, I found a case of the "array" option for an object array, so I'm interested in what others think.

typescript.js
TrevorKarjanis: In turns out, interfaces are already enforced by [@typescript-eslint/consistent-type-definitions](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-definitions.md), so I have a new proposal!

**Proposal**
Leave this rule unconfigured.

typescript.js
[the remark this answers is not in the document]
TrevorKarjanis: **Proposal**
```typescript
'@typescript-eslint/no-extraneous-class': ['error', { allowWithDecorator: true, allowStaticOnly: true }]
```

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/prefer-enum-initializers': 'error',
+        '@typescript-eslint/prefer-for-of': 'error',
+        '@typescript-eslint/prefer-function-type': 'error',
+        '@typescript-eslint/prefer-literal-enum-member': 'error',
+        '@typescript-eslint/prefer-optional-chain': 'error',
+        '@typescript-eslint/prefer-ts-expect-error': 'error',
+        // Available in newer typescript-eslint version
+        // '@typescript-eslint/sort-type-union-intersection-members': 'error',
+        '@typescript-eslint/type-annotation-spacing': 'error',
+        '@typescript-eslint/typedef': 'error',
```
TrevorKarjanis, line 58: [@typescript-eslint/typedef](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/typedef.md)
This rule doesn't do anything by default, and describes why it wouldn't be used in its description. I agree with its sentiment, and use implicit types where they are not required. I do, however, always define types for function and method parameters, because this is important for documenting their usage. Unfortunately, even this case is problematic with default parameters.

**Proposal**
Leave this rule unconfigured.

TrevorKarjanis (changes requested):

typescript.js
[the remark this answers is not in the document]
jattasNI: I was personally a bit annoyed when I saw this rule causing errors (in the O+ codebase they use a mix of both styles but the majority are `[]`). Both syntaxes are valid and most developers will have no trouble understanding either. But I can see some value in standardizing so I was ok to deal with the annoyance.

If we're going to enable the rule, I agree that the `[]` syntax feels more natural for simple arrays and the `<>` syntax feels more natural for arrays of compound types. 

But ultimately I don't feel too strongly about this one.

typescript.js
jattasNI: @LukasKall could you add a short comment to your proposal for no-implicit-any-catch explaining why we chose to turn it off (just to save Milan the effort of summarizing your comment)?

typescript.js
jattasNI: Since Carson is OOO, here's a 
**Proposal**: Disable this rule
```js
/* If an enum is crossing a code boundary (being serialized to JSON for example) then you should initialize its values. But in most other cases code shouldn't care about the values of enum items, so you can safely leave them implicit. */
'@typescript-eslint/prefer-enum-initializers': 'off'
```

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+
+        // None
+
+        /*
+            TypeScript rules outside of the recommended configuration
+        */
+        '@typescript-eslint/array-type': 'error',
+        '@typescript-eslint/ban-tslint-comment': 'error',
+        '@typescript-eslint/class-literal-property-style': 'error',
+        '@typescript-eslint/consistent-indexed-object-style': 'error',
```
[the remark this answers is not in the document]
jattasNI: @TrevorKarjanis are you suggesting that we disable `consistent-indexed-object-style` to allow both syntaxes? If so could you make a proposal for voting?

Or if you're just seeking feedback, here's mine: no strong opinion on this one. Personally I find the Record syntax to be a bit more intuitive, but I haven't seen either pattern enough to think that a lint rule will have much impact either way

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/no-unnecessary-type-constraint': 'error',
+        '@typescript-eslint/no-unused-vars-experimental': 'error',
+        '@typescript-eslint/prefer-enum-initializers': 'error',
+        '@typescript-eslint/prefer-for-of': 'error',
+        '@typescript-eslint/prefer-function-type': 'error',
+        '@typescript-eslint/prefer-literal-enum-member': 'error',
+        '@typescript-eslint/prefer-optional-chain': 'error',
+        '@typescript-eslint/prefer-ts-expect-error': 'error',
+        // Available in newer typescript-eslint version
+        // '@typescript-eslint/sort-type-union-intersection-members': 'error',
```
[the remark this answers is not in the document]
mure: It looks like this rule has an autofixer. Personally, I like having things alphabetized, but only if I don't have to do it manually. Also, don't we enforce alphabetization of imports?

typescript.js
mure: I'm unable to get the `allowInGenericTypeArguments` config to work correctly (as far as I understand it). I've filed an issue here: https://github.com/typescript-eslint/typescript-eslint/issues/3143

typescript.js
bradzacher: > I can't imagine a scenario where this would matter unless you are directly referencing an enum's value

The scenario it matters is when you serialise your enum values somewhere (eg to a file, or to a database).
Without explicit initialisers if you change the order (/ add new members) after the initial serialisation then your serialised data is no longer consistent and now carries a new meaning.

---

Another usecase to think of is that without an initialiser TS will set the first enum member to have value `0`.

Which means that the following code:
```ts
enum Foo {
  A,
  B,
}

function test(arg?: Foo | null) {
  console.log('is arg falsey', !arg);
}

test();          // is arg falsey true
test(undefined); // is arg falsey true
test(null);      // is arg falsey true
test(Foo.A);     // is arg falsey true
test(Foo.B);     // is arg falsey false
```

This can easily cause unintended bugs if your codebase uses loose boolean checks.

If you always use explicit initialisers it means that you can only run into this case if you explicitly set a member to a falsey value.
You can also get around this by using a rule like `@typescript-eslint/strict-boolean-expressions`.

typescript.js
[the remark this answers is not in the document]
jattasNI: @rajsite 🚀

typescript.js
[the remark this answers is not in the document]
jattasNI: @rajsite this is awaiting your vote

typescript.js
jattasNI: @rajsite This is awaiting your vote

typescript.js
jattasNI: @rajsite vote already

typescript.js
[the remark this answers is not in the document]
jattasNI: We spoke about this in person. Decision is to allow only the square bracket syntax. @TrevorKarjanis will write a comment explaining our rationale.

typescript.js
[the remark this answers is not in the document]
jattasNI: We spoke in person and agreed to enforce the index-signature style. @TrevorKarjanis will create a comment explaining our rationale.

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/array-type': 'error',
+        '@typescript-eslint/ban-tslint-comment': 'error',
+        '@typescript-eslint/class-literal-property-style': 'error',
+        '@typescript-eslint/consistent-indexed-object-style': 'error',
+        '@typescript-eslint/consistent-type-assertions': 'error',
+        '@typescript-eslint/consistent-type-definitions': 'error',
+        '@typescript-eslint/consistent-type-imports': 'error',
+        '@typescript-eslint/explicit-function-return-type': 'error',
+        '@typescript-eslint/explicit-member-accessibility': 'error',
+        '@typescript-eslint/member-delimiter-style': 'error',
```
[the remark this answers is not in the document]
jattasNI: @rajsite this is awaiting your vote

typescript.js
jattasNI: @rajsite 🚀 (unless you vote otherwise)

typescript.js
```diff
@@ -2,79 +2,60 @@ module.exports = {
...
+        '@typescript-eslint/member-ordering': 'error',
+        '@typescript-eslint/method-signature-style': 'error',
+        '@typescript-eslint/no-confusing-non-null-assertion': 'error',
+        '@typescript-eslint/no-dynamic-delete': 'error',
```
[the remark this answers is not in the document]
jattasNI: In our meeting today we discussed banning `delete` altogether, not just dynamic delete. If we want to do that, we'd use the `no-restricted-syntax` rule and turn this one off since it would be redundant.

**Proposal** 
ESLint rule
```js
/* The delete operator is most commonly used when using an object as a map. Use the ES Map data structure instead. */
"no-restricted-syntax": ["error", 
    {
        "selector": "UnaryExpression[operator='delete']",
        "message": "The 'delete' operator is not allowed. If using an object keys as a map, use the ES 'Map' data structure instead."
    }
]
```
TypeScript rule
```js
/* This rule is unnecessary because delete is banned via 'no-restricted-syntax'
"@typescript-eslint/no-dynamic-delete": "off"
```

typescript.js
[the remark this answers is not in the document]
TrevorKarjanis: **Proposal**
```JavaScript
/*
    Prefer the array straight bracket syntax over generics in all cases.
*/
'@typescript-eslint/array-type': 'error'
```

typescript.js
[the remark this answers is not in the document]
TrevorKarjanis: **Proposal**
```JavaScript
/*
    Prefer the index-signature syntax over generics in all cases.
*/
'@typescript-eslint/array-type': 'error'
```

typescript.js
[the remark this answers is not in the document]
TrevorKarjanis: I don't think people care about as much about imports, because there are so many of them and there is less desire to organize in any other way than third-party and alphabetically. Even with the autofixer, I could see non-alphabetizers getting annoyed and fighting with this rule. I don't always alphabetize unions myself, because I frequently place the more common types/values first.

typescript.js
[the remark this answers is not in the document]
LukasKall: I guess you don't mean `array-type` in the case, but instead
```
'@typescript-eslint/consistent-indexed-object-style': ['error', 'index-signature']
```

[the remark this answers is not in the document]
TrevorKarjanis: Thanks, I copied the comment from above and forgot to change it. You now have to proposal.

typescript.js
rajsite: Having `'@typescript-eslint/no-type-alias': 'off'` makes sense. The docs are not clear what the default configuration for `@typescript-eslint/consistent-type-definitions` is.

Does the proposal imply requiring interfaces so a configuration of `"@typescript-eslint/consistent-type-definitions": ["error", "interface"]`?

typescript.js
rajsite: Based on the issue discussion it sounds like the behavior described is not a false positive so we may need a source change to support this rule. I'll mark this as resolved.

typescript.js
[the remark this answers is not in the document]
rajsite: @mure do you have a vote for this rule?

typescript.js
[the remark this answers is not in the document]
rajsite: @mure have a vote for the latest proposal?

typescript.js
TrevorKarjanis: Yes, I guess I figured we would get to it, but it doesn't look like it's in the recommended set.

jattasNI (approved):

TrevorKarjanis (approved):

LukasKall (approved):

rajsite:
@mure I'll go ahead and merge this since we have most reviewers. If you have feedback we can address it in a follow-up PR.

PR #20, opened 2021-03-15
Enable TypeScript Rules Requiring Type Checking

rajsite:
The goal of this PR is to evaluate rules used in `typescript-requiring-type-checking.js`.

To test locally, you can add the dependencies to a project with the instructions in the current published [README](https://github.com/ni/javascript-styleguide#installation) as it uses the latest dependency versions.

**Note:** The dependency versions have changed since the last PR so make sure you are up to date.

Download the zip for the build artifact (see below) and extract the contents (ie the single tgz file). **Note:** you do not need to extract the tgz file itself. With the tgz file on disk run the following command:

`npm install -D <path_to_tgz_file>`

This PR also makes the following changes to review:
- Updates to the README.md

Build Artifacts:
- [Initial PR](https://github.com/ni/javascript-styleguide/actions/runs/654563358)

**NOTE: due to a mistake of deleting the source branch of this PR this PR will not be merged and instead the following branch will be in a follow-up PR: https://github.com/rajsite/javascript-styleguide/tree/enable-requires-typechecking-complete**

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Extend `@ni/eslint-config/typescript-requiring-type-checking` in a separate ESLi
 
 ```json
 {
-    "extends": ["@ni/eslint-config/typescript-requiring-type-checking"],
+    "extends": "@ni/eslint-config/typescript-requiring-type-checking",
     "parserOptions": {
         "project": "tsconfig.json"
     }
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Welcome to NI's JavaScript and TypeScript linter rules for [ESLint](https://esli
 
 ## Installation
 
-Install @ni/eslint-config and its peer dependencies.
+Install `@ni/eslint-config` and its peer dependencies.
 
 Use [npm view](https://docs.npmjs.com/cli/view.html) to list the correct versions of each package to install yourself.
 
@@ -24,7 +24,7 @@ npx install-peerdeps --dev @ni/eslint-config
 
 ### JavaScript
 
-Extend @ni in the ESLint configuration.
+Extend `@ni` in the default ESLint configuration, ie `.eslintrc.json`.
 
 ```json
 {
@@ -34,12 +34,21 @@ Extend @ni in the ESLint configuration.
 
 ### TypeScript
 
-Extend @ni/eslint-config/typescript in the ESLint configuration. Configure the @typescript-eslint plugin and the project's TypeScript configuration.
+Extend `@ni/eslint-config/typescript` in the default ESLint configuration. The default `.eslintrc.json` used for development should only extend TypeScript rules and not the TypeScript rules requiring type checking.
 
 ```json
 {
     "extends": "@ni/eslint-config/typescript",
-    "plugins": ["@typescript-eslint"],
+}
+```
+
+### TypeScript Rules Requiring Type Checking
+
+Extend `@ni/eslint-config/typescript-requiring-type-checking` in a separate ESLint configuration. Configure the project's TypeScript configuration. Due to the longer execution time, the separate `.eslintrc-requires-type-checking.json` should only run from an explicit `npm run lint` command invoked by developers and on CIs.
+
+```json
+{
+    "extends": ["@ni/eslint-config/typescript-requiring-type-checking"],
     "parserOptions": {
         "project": "tsconfig.json"
     }
--- a/lib/typescript-extensions-requiring-type-checking.js
+++ b/lib/typescript-extensions-requiring-type-checking.js
@@ -1,23 +1,23 @@
 module.exports = {
     rules: {
         // Defined by Airbnb
-        // 'dot-notation': 'off',
-        // '@typescript-eslint/dot-notation': ['error', { allowKeywords: true }],
+        'dot-notation': 'off',
+        '@typescript-eslint/dot-notation': ['error', { allowKeywords: true }],
 
         // Defined by Airbnb
-        // 'no-implied-eval': 'off',
-        // '@typescript-eslint/no-implied-eval': 'error',
+        'no-implied-eval': 'off',
+        '@typescript-eslint/no-implied-eval': 'error',
 
         // Defined by Airbnb
-        // 'no-throw-literal': 'off',
-        // '@typescript-eslint/no-throw-literal': 'error',
+        'no-throw-literal': 'off',
+        '@typescript-eslint/no-throw-literal': 'error',
 
         // Defined by Airbnb
-        // 'require-await': 'off',
-        // '@typescript-eslint/require-await': 'off',
+        'require-await': 'off',
+        '@typescript-eslint/require-await': 'off',
 
         // Defined by Airbnb
-        // 'no-return-await': 'off',
-        // '@typescript-eslint/return-await': 'error',
+        'no-return-await': 'off',
+        '@typescript-eslint/return-await': 'error',
     }
 };
--- a/typescript-requiring-type-checking.js
+++ b/typescript-requiring-type-checking.js
@@ -15,23 +15,23 @@ module.exports = {
         /*
             TypeScript rules outside of the recommended configuration
         */
-        // '@typescript-eslint/naming-convention': 'error',
-        // '@typescript-eslint/no-base-to-string': 'error',
-        // '@typescript-eslint/no-confusing-void-expression': 'error',
-        // '@typescript-eslint/no-unnecessary-boolean-literal-compare': 'error',
-        // '@typescript-eslint/no-unnecessary-condition': 'error',
-        // '@typescript-eslint/no-unnecessary-qualifier': 'error',
-        // '@typescript-eslint/no-unnecessary-type-arguments': 'error',
-        // '@typescript-eslint/non-nullable-type-assertion-style': 'error',
-        // '@typescript-eslint/prefer-includes': 'error',
-        // '@typescript-eslint/prefer-nullish-coalescing': 'error',
-        // '@typescript-eslint/prefer-readonly': 'error',
-        // '@typescript-eslint/prefer-readonly-parameter-types': 'error',
-        // '@typescript-eslint/prefer-reduce-type-parameter': 'error',
-        // '@typescript-eslint/prefer-string-starts-ends-with': 'error',
-        // '@typescript-eslint/promise-function-async': 'error',
-        // '@typescript-eslint/require-array-sort-compare': 'error',
-        // '@typescript-eslint/strict-boolean-expressions': 'error',
-        // '@typescript-eslint/switch-exhaustiveness-check': 'error'
+        '@typescript-eslint/naming-convention': 'error',
+        '@typescript-eslint/no-base-to-string': 'error',
+        '@typescript-eslint/no-confusing-void-expression': 'error',
+        '@typescript-eslint/no-unnecessary-boolean-literal-compare': 'error',
+        '@typescript-eslint/no-unnecessary-condition': 'error',
+        '@typescript-eslint/no-unnecessary-qualifier': 'error',
+        '@typescript-eslint/no-unnecessary-type-arguments': 'error',
+        '@typescript-eslint/non-nullable-type-assertion-style': 'error',
+        '@typescript-eslint/prefer-includes': 'error',
+        '@typescript-eslint/prefer-nullish-coalescing': 'error',
+        '@typescript-eslint/prefer-readonly': 'error',
+        '@typescript-eslint/prefer-readonly-parameter-types': 'error',
+        '@typescript-eslint/prefer-reduce-type-parameter': 'error',
+        '@typescript-eslint/prefer-string-starts-ends-with': 'error',
+        '@typescript-eslint/promise-function-async': 'error',
+        '@typescript-eslint/require-array-sort-compare': 'error',
+        '@typescript-eslint/strict-boolean-expressions': 'error',
+        '@typescript-eslint/switch-exhaustiveness-check': 'error'
     }
 };


README.md
```diff
@@ -34,12 +34,21 @@ Extend @ni in the ESLint configuration.
 
 ### TypeScript
 
-Extend @ni/eslint-config/typescript in the ESLint configuration. Configure the @typescript-eslint plugin and the project's TypeScript configuration.
+Extend `@ni/eslint-config/typescript` in the default ESLint configuration. The default `.eslintrc.json` used for development should only extend TypeScript rules and not the TypeScript rules requiring type checking.
 
 ```json
 {
     "extends": "@ni/eslint-config/typescript",
-    "plugins": ["@typescript-eslint"],
```
rajsite, line 42: I removed the plugins config from the root README as it is already in the [typescript-eslint base](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/base.ts#L8) and it seems to work without it.

Let me know if we should add it back. The [getting started docs](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/README.md#configuration) seem to indicate configuring both `plugins` and `parser` is needed but it does not seem that way locally and I think it's because they are configured in base.

jattasNI: I didn't have any problems when I removed `plugins` from my config files.

typescript-requiring-type-checking.js
```diff
@@ -15,23 +15,23 @@ module.exports = {
         /*
             TypeScript rules outside of the recommended configuration
         */
```
rajsite, line 17: test

typescript-requiring-type-checking.js
```diff
@@ -15,23 +15,23 @@ module.exports = {
         /*
```
rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/await-thenable': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/no-floating-promises': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/no-for-in-array': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/no-implied-eval'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/no-misused-promises': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/no-unnecessary-type-assertion': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/no-unsafe-assignment': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/no-unsafe-call': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/no-unsafe-member-access': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/no-unsafe-return': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/prefer-regexp-exec': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/require-await': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/restrict-plus-operands': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/restrict-template-expressions': 'error'`

rajsite, line 15: Discussion for recommended rule: `'@typescript-eslint/unbound-method': 'error'`

README.md
```diff
@@ -34,12 +34,21 @@ Extend @ni in the ESLint configuration.
...
+}
+```
+
+### TypeScript Rules Requiring Type Checking
+
+Extend `@ni/eslint-config/typescript-requiring-type-checking` in a separate ESLint configuration. Configure the project's TypeScript configuration. Due to the longer execution time, the separate `.eslintrc-requires-type-checking.json` should only run from an explicit `npm run lint` command invoked by developers and on CIs.
```
rajsite, line 47: Think we should recommend a git pre commit hook or a git pre push hook here? Might decrease developer frustration when forgetting to run before pushing to PRs for PR checks to run. Could increase frustration storing changes on temporary branches. I haven't really used them myself.

typescript-requiring-type-checking.js
jattasNI: I ran into several errors with [@typescript-eslint/no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md) that are worth discussing.

1. In Jasmine tests, assert functions return `Promise<void>` so lines like this violate the rule: `expect(serialized.widgets).toBeDefined();`.
   - possible mitigation: turn this off for test code
   - ugly mitigation from [StackOverflow](https://stackoverflow.com/questions/56938854/how-to-ignore-jasmine-promises-when-no-floating-promises-in-tslint-is-true): add the `void` keyword before every assert. `void expect(serialized.widgets).toBeDefined();`
1. Angular's `router.navigate()` function returns a `Promise<boolean>` that describes whether the navigation was successful but [none of NI's apps](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=router.navigate%20ext%3Ats&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7D&pageSize=25&result=DefaultCollection%2FDevCentral%2Fsrd-semi-ses-software%2FGBmain%2F%2FSpecManagement%2Fsrc%2Fapp%2Fauthentication%2Fauth.guard.ts) await or otherwise handle that promise.
1. In [production code](https://dev.azure.com/ni/DevCentral/_git/op-web-portal?path=%2FOptimalTest.WebPortal%2Fsrc%2Fapp%2Fapp.component.ts&version=GBmain&line=42&lineEnd=43&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents), calling an async function without awaiting it. This seems like a legit error and we should fix it by awaiting or explicitly saying we don't need it to complete (via `void`?)
1. [Some code](https://dev.azure.com/ni/DevCentral/_git/op-web-portal?path=%2FOptimalTest.WebPortal%2Fsrc%2Fapp%2Fdashboard-catalog%2Fcomponents%2Fdashboard-catalog%2Fdashboard-catalog.component.ts&version=GBmain&line=267&lineEnd=267&lineStartColumn=29&lineEndColumn=51&lineStyle=plain&_a=contents) handles a promise via `then()` but only handles the `resolve` case but not the `reject`. Would our recommendation be to always provide a `reject` callback?

jattasNI: I saw the following failures with [restrict-plus-operands](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/restrict-plus-operands.md) that are worth discussing:
1. String formatting like 
   - `numericValue1 + numericValue2 + 'px'` 
   - `marker.value + '%` 
   - `'Can't select more than ` + number + ' items.'`
   - `urlString + '/' + idNumber`
   The rule suggests replacing these with template literals but is that sufficiently more readable that it's worth enforcing?
2. Adding two variables with `any` type that were obviously numbers. This seems like a legit failure.

I think I like this rule but since I hit a decent number of failures (>100) I thought it would be good to discuss.

jattasNI: Point of clarification on 1. since Milan was asking about it offline. The functions that return a promise are the "matchers" (`toBeDefined()`, `toEqual()`, etc). This [behavior](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0b5bfba2994c91a099cd5bcfd984f6c4c39228e5/types/jasminewd2/index.d.ts) is specific to [jasminewd2](https://www.npmjs.com/package/jasminewd2), which is a version of Jasmine used by Protractor (and used by default in Angular apps) with async enhancements for Web Driver.

jattasNI: I noticed that [restrict-template-expressions](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/restrict-template-expressions.md) would trigger in the default case of the following code, saying "Invalid type "never" of template literal expression":
```ts
enum Foo { First, Second }

someFunc(x: Foo) {
    switch (x) {
        case Foo.First: return 'first';
        case Foo.Second: return 'second';
        default: throw new Error(`Invalid enum value ${x}`);
    }
}
```
What would be our recommended fix for this error?

jattasNI: Lots of interesting violations of [unbound-method](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/unbound-method.md) in the Web Portal codebase with the error "Avoid referencing unbound methods which may cause unintentional scoping of `this`":
1. [Using lodash](https://dev.azure.com/ni/DevCentral/_git/op-web-portal?path=%2FOptimalTest.WebPortal%2Fsrc%2Fapp%2Fbars-and-lines-widget%2Fcomponents%2Fbars-and-lines-component%2Fbars-and-lines.component.ts&version=GBmain&line=234&lineEnd=235&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents): `items = _.uniqWith(selectionObj.selection, _.isEqual);`. I'm not familiar enough with lodash or scoping rules to know if this is a true problem or how to fix it.
2. [Passing an instance method as a callback](https://dev.azure.com/ni/DevCentral/_git/op-web-portal?path=%2FOptimalTest.WebPortal%2Fsrc%2Fapp%2Fboxplot-widget-new%2Fservices%2Fboxplot-options-creator.service.ts&version=GBmain&line=534&lineEnd=535&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) to a sort function. `serie.data.sort(this.sortPoints);` The fix is to change `sortPoints` to `static` (it doesn't use `this`).
3. [Calling `removeEventListener` with an instance method](https://dev.azure.com/ni/DevCentral/_git/op-web-portal?path=%2FOptimalTest.WebPortal%2Fsrc%2Fapp%2Fchart-widget-common%2Fservices%2Flabel-creator.service.ts&version=GBmain&line=12&lineEnd=13&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents): `xAxisLabels[i].removeEventListener('mouseout', this.hideTooltip);` [MDN has a detailed description of this behavior](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_value_of_this_within_the_handler). Is this code doing anything wrong?
4. [Test code which calls a helper function](https://dev.azure.com/ni/DevCentral/_git/op-web-portal?path=%2FOptimalTest.WebPortal%2Fsrc%2Fapp%2Fcore%2Fservices%2Fstream-response-extraction-factory.service.spec.ts&version=GBmain&line=32&lineEnd=32&lineStartColumn=16&lineEndColumn=28&lineStyle=plain&_a=contents) defined inline which doesn't use `this`.
5. [Passing an instance method](https://dev.azure.com/ni/DevCentral/_git/op-web-portal?path=%2FOptimalTest.WebPortal%2Fsrc%2Fapp%2Fhistogram-widget%2Fservices%2Fhistogram-data.service.ts&version=GBmain&line=76&lineEnd=77&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) as a callback when it [does use `this`](https://dev.azure.com/ni/DevCentral/_git/op-web-portal?path=%2FOptimalTest.WebPortal%2Fsrc%2Fapp%2Fcore%2Fservices%2Fbase-http.service.ts&version=GBmain&line=20&lineEnd=21&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents). If `this` isn't passed correctly then this could be catching a bug.

It's clearly a gap in my own knowledge that I can't tell if these are correct or not but I wonder if the rule will help developers in the same boat as me to educate themselves or just bewilder/annoy them.

jattasNI: The fixer for this rule didn't work for me. It left extra `<` character when removing a `<Conversion>`. I filed a bug but I'm curious if anyone else sees this. https://github.com/typescript-eslint/typescript-eslint/issues/3196

rajsite: Recently read a blog called [Don't use functions as callbacks unless they're designed for it](https://jakearchibald.com/2021/function-callback-risks/) that gives more examples that seem to be in favor of this rule. I know I've asked folks to use lambdas in cases where arguments don't line up or for binding a `this` context in SystemLink PRs.

lib/typescript-extensions-requiring-type-checking.js
```diff
@@ -1,23 +1,23 @@
...
-        // 'dot-notation': 'off',
-        // '@typescript-eslint/dot-notation': ['error', { allowKeywords: true }],
+        'dot-notation': 'off',
+        '@typescript-eslint/dot-notation': ['error', { allowKeywords: true }],
 
         // Defined by Airbnb
-        // 'no-implied-eval': 'off',
-        // '@typescript-eslint/no-implied-eval': 'error',
+        'no-implied-eval': 'off',
+        '@typescript-eslint/no-implied-eval': 'error',
```
jattasNI, line 9: I got two errors from [no-implied-eval](https://github.com/typescript-eslint/typescript-eslint/blob/498f397ff3898dde631f37311615b555f38a414e/packages/eslint-plugin/docs/rules/no-implied-eval.md) that seem to me to be mistaken. In both cases the code is defining a function then calling `bind` on it before passing it to `setTimeout`. The rule is supposed to trigger if you pass a string to `setTimeout` but it looks like `bind` produces a function, not a string. Does anyone know why this should be an error?
```ts
setTimeout(function () {
    this.modalFilters.show();
}.bind(this), 0);
```

rajsite: Not sure why the bind call would trigger no-implied-eval. Their examples even include something pretty similar as a correct usage example:
```ts
setTimeout(foo.fn.bind(this), 100);
```

But this case is probably better written with an arrow function to bind the this context as:
```ts
setTimeout(() => this.modalFilters.show(), 0);
```

typescript-requiring-type-checking.js
jattasNI: I agree, it's a good read. One thing that might help developers who hit this error would be if the error message guided them towards arrow functions. I think it's possible for us to customize the error message in our rule config, so if people agree that's a good idea then I can create a proposal that does that.

typescript-requiring-type-checking.js
```diff
@@ -15,23 +15,23 @@ module.exports = {
...
+        '@typescript-eslint/no-confusing-void-expression': 'error',
+        '@typescript-eslint/no-unnecessary-boolean-literal-compare': 'error',
+        '@typescript-eslint/no-unnecessary-condition': 'error',
+        '@typescript-eslint/no-unnecessary-qualifier': 'error',
+        '@typescript-eslint/no-unnecessary-type-arguments': 'error',
+        '@typescript-eslint/non-nullable-type-assertion-style': 'error',
+        '@typescript-eslint/prefer-includes': 'error',
+        '@typescript-eslint/prefer-nullish-coalescing': 'error',
+        '@typescript-eslint/prefer-readonly': 'error',
+        '@typescript-eslint/prefer-readonly-parameter-types': 'error',
```
LukasKall, line 29: **[@typescript-eslint/prefer-readonly-parameter-types](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-readonly-parameter-types.md)**

This rule doesn't give much additional value since we have already enabled [no-param-reassign](https://eslint.org/docs/rules/no-param-reassign) which covers this.

**Proposal**
```
/*
    This is already covered by `no-param-reassign`.
*/
'@typescript-eslint/prefer-readonly-parameter-types': 'off'
```

lib/typescript-extensions-requiring-type-checking.js
jattasNI: General agreement that we should keep this rule. We have other rules that should enforce the arrow syntax preference. Jesse can file a bug to typescript-eslnt about the bind behavior.

typescript-requiring-type-checking.js
jattasNI: We should check if we have a rule that ensures `switch` statements are fully covering every enum value. If so then the recommendation would be not to provide a `default` case.

In cases where we don't know all the enum values (if coming from an external library) then it would be appropriate to cast the value in the template literal to tell the compiler what's going on (if necessary...in that case it might not be `never` type)

jattasNI: General agreement at today's meeting that this is a good rule but I'll leave this thread open in case people have more thoughts after adopting it in other repos.

jattasNI: General consensus at our meeting today was that the rule is providing some value in cases 2, 3, and 4. Case 1 is unfortunate and we might provide guidance that the rule can be disabled for test code that uses `jasminewd2`. I'll leave this thread open for now but since it was moderately controversial it should be resolved by voting on a proposal.

jattasNI: We started discussing this at today's meeting but didn't complete. Most of these cases should be solved with arrow functions, but we should talk about `1` in more detail. Interesting note is that the code in `3` is actually wrong: `removeEventListener` is not removing the same callback that `addEventListener` added.

typescript-requiring-type-checking.js
mure: Although the end result is very similar, requiring the 'readonly' keyword is more explicit than just preventing parameter reassign.
From the description:
>Whilst it's easy to implicitly remember to not modify function arguments, explicitly typing arguments as readonly provides clear contract to consumers.

I don't have much of an opinion on this rule, but it's worth discussing as a group.

typescript-requiring-type-checking.js
```diff
@@ -15,23 +15,23 @@ module.exports = {
...
-        // '@typescript-eslint/prefer-nullish-coalescing': 'error',
-        // '@typescript-eslint/prefer-readonly': 'error',
-        // '@typescript-eslint/prefer-readonly-parameter-types': 'error',
-        // '@typescript-eslint/prefer-reduce-type-parameter': 'error',
-        // '@typescript-eslint/prefer-string-starts-ends-with': 'error',
-        // '@typescript-eslint/promise-function-async': 'error',
-        // '@typescript-eslint/require-array-sort-compare': 'error',
-        // '@typescript-eslint/strict-boolean-expressions': 'error',
-        // '@typescript-eslint/switch-exhaustiveness-check': 'error'
+        '@typescript-eslint/naming-convention': 'error',
```
mure, line 18: This rule is erroring on Enum's _not_ being camel case, which is against the norm that I've seen in our TS codebases. 
>`Enum Member name 'Create' must match one of the following formats: camelCase` 

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md

There's a very extensive config for this rule that we probably want to go over in a meeting.

typescript-requiring-type-checking.js
mure: Should we make `ignoreStatic` true for this rule? An example using Angular's reactive forms:
```
documentForm = new FormGroup({
    name: new FormControl('', [
        Validators.required,
        DocumentSettingsComponent.invalidCharacters
    ]),
    description: new FormControl('', [
        DocumentSettingsComponent.invalidCharacters
    ]),
    workspace: new FormControl('', Validators.required)
});
```
By default, this rule will error on the `Validators.required` reference (which is static and coming directly from Angular's API).

typescript-requiring-type-checking.js
```diff
@@ -15,23 +15,23 @@ module.exports = {
...
+        '@typescript-eslint/no-base-to-string': 'error',
+        '@typescript-eslint/no-confusing-void-expression': 'error',
```
mure, line 20: This rule requires braces around arrow functions that return void, which I think is unnecessary in many cases.
Example from an observable chain:
 ```
this.propertyChange$.pipe(
    filter((): boolean => this.canLoadIframe()),
    filter((changes: SimpleChanges): boolean => this.changeRequiresReload(changes)),
    switchMap((): Observable<Document> => this.getDocument()),
    tap((document: Document): Document => this.document = document),
    tap(() => this.updateIframe())
).subscribe()
```

This rule will error on `tap(() => this.updateIframe())` and wants it to be something like:
```
this.propertyChange$.pipe(
    filter((): boolean => this.canLoadIframe()),
    filter((changes: SimpleChanges): boolean => this.changeRequiresReload(changes)),
    switchMap((): Observable<Document> => this.getDocument()),
    tap((document: Document): Document => this.document = document),
    tap(() => {
        this.updateIframe()
    })
).subscribe()
```

**Proposal**
```
"@typescript-eslint/no-confusing-void-expression": [
    "error",
    { "ignoreArrowShorthand": true }
]
```

File: typescript-requiring-type-checking.js (hunk 3 above)
Comment (TrevorKarjanis) [reply typescript-requiring-type-checking.js:15]: Should we just enable [no-explicit-any](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-explicit-any.md) instead?

Comment (TrevorKarjanis) [reply typescript-requiring-type-checking.js:15]: Should we just enable [no-explicit-any](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-explicit-any.md) instead?

Comment (TrevorKarjanis) [reply typescript-requiring-type-checking.js:15]: Should we just enable [no-explicit-any](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-explicit-any.md) instead?

Comment (TrevorKarjanis) [reply typescript-requiring-type-checking.js:15]: Should we just enable [no-explicit-any](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-explicit-any.md) instead?

Comment (TrevorKarjanis) [reply typescript-requiring-type-checking.js:15]: [@typescript-eslint/prefer-regexp-exec](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-regexp-exec.md)
The performance gain proposed here is negligible in most cases. People prefer to use match with strings, because it's more intuitive.

Comment (TrevorKarjanis) [reply typescript-requiring-type-checking.js:15]: This rule is disabled in the extensions configuration.
**Proposal**
Leave this rule un-configured.

Comment (TrevorKarjanis) [reply typescript-requiring-type-checking.js:15]: This rule is enabled in the extensions.
**Proposal**
Leave this rule un-configured.

Comment (TrevorKarjanis) [reply typescript-requiring-type-checking.js:15]: [switch-exhaustiveness-check](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/switch-exhaustiveness-check.md) will check enums and enum like types in switch statements. We already enabled [default-case](https://eslint.org/docs/rules/default-case), so excluding the default case will require a comment.

File: typescript-requiring-type-checking.js [hunk 9]
```diff
@@ -15,23 +15,23 @@ module.exports = {
...
+        '@typescript-eslint/no-unnecessary-boolean-literal-compare': 'error',
```
TrevorKarjanis, line 21: [@typescript-eslint/no-unnecessary-boolean-literal-compare](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unnecessary-boolean-literal-compare.md)
"Warning: Do not use this rule when strictNullChecks is disabled. ESLint is not able to distinguish between false and undefined or null values. This can cause unintended code changes when using autofix."

typescript-requiring-type-checking.js
jattasNI: Discussed in person: we think it's ok to have both no-explicit-any and this rule enabled so let's leave the rule configured as it is in the recommended config.

jattasNI: Discussed in person: we think it's ok to have both no-explicit-any and this rule enabled so let's leave the rule configured as it is in the recommended config.

jattasNI: Discussed in person. We are OK with Trevor's proposal to disable this rule, though @rajsite is going to do research into side effects of calling `match`. We can use Trevor's comment as the explanation.

jattasNI: The config is already arranged so that the extensions takes priority over the recommended rule, so we don't need to do anything here.

jattasNI: We discussed in person. General agreement that we should `ignoreStatic` even though there are some cases where static methods might be using `this`. The annoyance of having to disable the rule for Carson's case outweighs the benefit in other cases.

README.md
jattasNI: Discussed in person. General consensus is that it would be annoying to prevent local commits due to type errors.

typescript-requiring-type-checking.js
jattasNI: We agreed in person that this rule is too aggressive so we should disable it. It's not precisely the same as `no-param-reassign` so here's a new explanation:
```
/*
    This rule requires adding `readonly` keywords to all levels of non-primitive parameters in order to prevent any part of the parameter from being mutated. While the spirit of preferring not to mutate inputs is correct, requiring it to be encoded via `readonly` is overly burdensome.
*/
'@typescript-eslint/prefer-readonly-parameter-types': 'off'
```

typescript-requiring-type-checking.js
rajsite: Example of how match has side-effects (mutates properties on the RegExp constructor):
image

typescript-requiring-type-checking.js
```diff
@@ -15,23 +15,23 @@ module.exports = {
...
+        '@typescript-eslint/no-unnecessary-condition': 'error',
+        '@typescript-eslint/no-unnecessary-qualifier': 'error',
+        '@typescript-eslint/no-unnecessary-type-arguments': 'error',
+        '@typescript-eslint/non-nullable-type-assertion-style': 'error',
+        '@typescript-eslint/prefer-includes': 'error',
+        '@typescript-eslint/prefer-nullish-coalescing': 'error',
```
TrevorKarjanis, line 27: [@typescript-eslint/prefer-nullish-coalescing](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md)
Actually, the examples for this one would be bugs. I don't use the nullish operator specifically to check for empty string. My argument against this one is similar to strict-boolean-expressions below. Non-boolean expressions are useful/awesome and very commonly used.

typescript-requiring-type-checking.js
jattasNI: The bug was fixed in 4.21.0 which released a couple days ago so we should upgrade from our current 4.17.0 at some point. https://github.com/typescript-eslint/typescript-eslint/blob/master/CHANGELOG.md

typescript-requiring-type-checking.js
```diff
@@ -15,23 +15,23 @@ module.exports = {
...
+        '@typescript-eslint/prefer-readonly': 'error',
+        '@typescript-eslint/prefer-readonly-parameter-types': 'error',
+        '@typescript-eslint/prefer-reduce-type-parameter': 'error',
+        '@typescript-eslint/prefer-string-starts-ends-with': 'error',
+        '@typescript-eslint/promise-function-async': 'error',
+        '@typescript-eslint/require-array-sort-compare': 'error',
+        '@typescript-eslint/strict-boolean-expressions': 'error',
```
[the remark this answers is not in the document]
rajsite: And as I demonstrated before, here is where I come in with a [very recent example](https://dev.azure.com/ni/DevCentral/_git/op-web-portal/pullrequest/147444) of where that bit us. ♥

README.md
rajsite: Was also tested by nicolas in web portal so marking resolved

typescript-requiring-type-checking.js
rajsite: Proposal: Leave the rule enabled and add a comment similar to "Angular projects with tests leveraging the `jasminewd2` library may choose to disable this rule for test code but should consider keeping it enabled for source."

rajsite: This case is specifically preventing use of for-in for Arrays and not Objects. My understanding of:
> For...in statements can be used when performance is a consideration as long as guard-for-in is enable

Was for performance considerations when accessing members of an object. I don't think we should ever use for..in for iterating across indexes of an array, particularly if we are citing performance as the reason.

When targeting performance arrays should be iterated using three-clause for-loops `for (let i = 0; i < arr.length; i++) {}`

Proposal: Do not disable the recommended rule

rajsite: As it sounds like the issue is only in the fixer and not in the rule we can do the version update in a follow-up PR.
Proposal: Do not disable the recommended rule.

jattasNI: We're all in agreement that this is a good rule. 
@rajsite will double check whether this applies only to protractor tests or also to Jasmine tests. If the former, no special comment is needed. If the latter, we should keep the comment and maybe add an example about how to suppress the rule for certain files.

jattasNI: We agreed on Milan's proposal in person.

jattasNI: We can close this thread with no action

jattasNI: This thread can be resolved.

jattasNI: @rajsite this config needs to be changed to turn on `ignoreStatic`

typescript-requiring-type-checking.js
jattasNI: At a minimum we should change enums to be `PascalCase` and not `YELLING_CASE` or `snake_case`, or `camelCase`. @mure will audit the remainder of the config options this week. If he doesn't find time, we can consider disabling this rule temporarily and revisiting this in another PR

typescript-requiring-type-checking.js
jattasNI: This has the votes needed

typescript-requiring-type-checking.js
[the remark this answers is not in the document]
jattasNI: We agreed that we should turn this rule off because we don't require strictNullChecks

typescript-requiring-type-checking.js
```diff
@@ -15,23 +15,23 @@ module.exports = {
...
-        // '@typescript-eslint/prefer-string-starts-ends-with': 'error',
-        // '@typescript-eslint/promise-function-async': 'error',
-        // '@typescript-eslint/require-array-sort-compare': 'error',
-        // '@typescript-eslint/strict-boolean-expressions': 'error',
-        // '@typescript-eslint/switch-exhaustiveness-check': 'error'
+        '@typescript-eslint/naming-convention': 'error',
+        '@typescript-eslint/no-base-to-string': 'error',
+        '@typescript-eslint/no-confusing-void-expression': 'error',
+        '@typescript-eslint/no-unnecessary-boolean-literal-compare': 'error',
+        '@typescript-eslint/no-unnecessary-condition': 'error',
```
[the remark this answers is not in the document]
jattasNI: We started talking about this in our meeting today but didn't get to the bottom of which compiler options are set for the examples. This could use more investigation.

lib/typescript-extensions-requiring-type-checking.js
jattasNI: I tried to create a reproducible case to file a bug but didn't get anywhere.
- wrapping the function in parens `(function () {...})` still fails in Web Portal
- copying the failing code into the typescript-eslint test suite passes
- using the arrow syntax also passes

Since we have a workaround I don't think this is high enough priority to investigate further.

typescript-requiring-type-checking.js
mure: **Proposal:**
To match what seems to be a standard convention, we should force enums to be PascalCase rather than camelCase.
Building on the default config:
```
{
  "@typescript-eslint/naming-convention": [
    "error",
  {
    selector: 'default',
    format: ['camelCase'],
    leadingUnderscore: 'allow',
    trailingUnderscore: 'allow',
  },
  {
    selector: 'variable',
    format: ['camelCase', 'UPPER_CASE'],
    leadingUnderscore: 'allow',
    trailingUnderscore: 'allow',
  },
  {
    selector: 'enumMember',
    format: ['PascalCase']
  },
  {
    selector: 'typeLike',
    format: ['PascalCase'],
  },
]}
```

Interestingly, there's an issue that points out this inconsistency, but it was closed (incorrectly I think): https://github.com/typescript-eslint/typescript-eslint/issues/3260

typescript-requiring-type-checking.js
rajsite: As a group decided having the rule and `no-explicit-any` makes sense. A [separate PR](https://github.com/ni/javascript-styleguide/pull/30) to elevate the recommended `no-explicit-any` rule from `warn` to `error`.

rajsite: As a group decided having the rule and `no-explicit-any` makes sense. A [separate PR](https://github.com/ni/javascript-styleguide/pull/30) to elevate the recommended `no-explicit-any` rule from `warn` to `error`.

rajsite: As a group decided having the rule and `no-explicit-any` makes sense. A [separate PR](https://github.com/ni/javascript-styleguide/pull/30) to elevate the recommended `no-explicit-any` rule from `warn` to `error`.

rajsite: As a group decided having the rule and `no-explicit-any` makes sense. A [separate PR](https://github.com/ni/javascript-styleguide/pull/30) to elevate the recommended `no-explicit-any` rule from `warn` to `error`.

typescript-requiring-type-checking.js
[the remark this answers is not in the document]
TrevorKarjanis: I don't think it demonstrates much to produce an even rarer example. That may better demonstrate a lack of testing and review, but I'll leave that argument be because we can't catch every bug. It certainly demonstrates less than all of the cases where we've used unstrict expressions to our benefit, especially checking if a method exists before executing it.

typescript-requiring-type-checking.js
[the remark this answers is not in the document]
TrevorKarjanis: Following up, strict null checks both [remove null and undefined from the domain](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#--strictnullchecks) of every type and [require a check](https://www.typescriptlang.org/tsconfig#strictNullChecks) or non-null assertion to provide a gaurantee.

typescript-requiring-type-checking.js
rajsite: @mure will update the comment to emphasize what changes from the default. We may consider re-opening / posting a new issue.

typescript-requiring-type-checking.js
rajsite: Looks like it effects both test sweets. This is the comment I added:

```js
        // Notes for recommended configuration: '@typescript-eslint/no-floating-promises': 'error'
        // The jasminewd2 library used by Angular applications will result in a significant number of floating promises.
        // Consider disabling the '@typescript-eslint/no-floating-promises' rule for tests that utilize jasmine expects.
```

rajsite: Marking as resolved as no recent discussion, we can reopen as needed.

lib/typescript-extensions-requiring-type-checking.js
rajsite: @jattasNI should we report a bug to typescript-eslint? The error does seem incorrect.

Marking the thread as resolved since we will not modify the current config due to the incorrect behavior.

typescript-requiring-type-checking.js
jattasNI: @jattasNI will verify that switch-exhaustiveness-check and default-case play well together. 

The restrict-template-expressions rule that triggered this discussion is fine.

typescript-requiring-type-checking.js
jattasNI: It's also worth investigating whether we can only override the areas that are changing rather than copying the base configuration.

We agree on the direction, @mure just needs to update the config with comments or less duplicate information.

typescript-requiring-type-checking.js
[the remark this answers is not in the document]
jattasNI: We agreed in person that this rule should be disabled by default because most of our codebases don't enable strict null checks. If the rule is turned on in those codebases it would always throw an error saying the rule is not supported. @rajsite this is ready to close.

typescript-requiring-type-checking.js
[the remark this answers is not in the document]
jattasNI: @TrevorKarjanis will test how this rule behaves when strict null checks are off. Ideally we'll block numbers from being treated as booleans. We might be able to set that by using allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing and allowNullableNumber but we only want to do that if we're confident it will behave correctly.

typescript-requiring-type-checking.js
jattasNI: @jattasNI to check if this rule conflicts with the rule that caused [Nicolas' PR](https://dev.azure.com/ni/DevCentral/_git/op-web-portal/pullrequest/155326) to remove instances of logical chaining.

In our in person discussion we were split between preferring || and ??, especially whether '' would typically be treated as invalid (in the same category as null or undefined) or valid (in the same category as non-empty string).

jattasNI: The rule in Nicolas' PR was **[no-unused-expressions](https://eslint.org/docs/rules/no-unused-expressions)** which reports errors for code like this: `this.selectionChangedSubscription && this.selectionChangedSubscription.unsubscribe();`. The recommendation in that case is to use optional chaining (`.?`): `this.selectionChangedSubscription?.unsubscribe();`

`prefer-nullish-coalescing` only fires for `||` whereas the pattern in `no-unused-expressions` was `&&` so I don't think they're in conflict.

I'll put up proposals for the two options under discussion: please vote to see where we all stand.

jattasNI: **Proposal** 
```
        /* When coalescing (providing a valid value to replace a possibly invalid value), 
         * prefer nullish coalescing (`??`) over logical OR (`||`)
         * as the former will only coalesce for nullish values (null or undefined) 
         * while the latter will coalesce for all falsy values (empty string, 0) which is more likely to be unintended.
         */
        '@typescript-eslint/prefer-nullish-coalescing': 'error',

File: typescript-requiring-type-checking.js (hunk 3 above)
Comment (jattasNI) [reply typescript-requiring-type-checking.js:15]: **Proposal** Keep `default-case` as `error` but turn off `switch-exhaustiveness-check`.
```
/* Always provide a `default` case in `switch` statements. If the default case is logically unreachable, throw an Error. If the default case should do nothing, add a comment explaining why to indicate that it is an intentional decision.*/
"default-case": "error"

/* This rule is made redundant by `default-case` so it will never fire
"@typescript-eslint/switch-exhaustiveness-check": "off"
```

Here's my rationale:
1. There are some patterns that only `default-case` can detect. For example, when exhaustiveness can't be calculated (switching on a `string` with any possible value), our ruleset should enforce having a `default` case to signal that the developer thought about all possible values.
2. `default-case` is a bit annoying when the compiler and `switch-exhaustiveness-check` detect that the cases are exhaustive because it forces an unnecessary default case. But even though the case is unreachable, it doesn't cause any compiler errors. And it gives us a place to add a `throw` to make it obvious when reading the code that it is unreachable.
3. Once `default-case` is on, `switch-exhaustiveness-check` would never fail because adding a `default` case will always make the `switch` exhaustive. It's not worth the time to enforce it so we should turn it off.

I created an [example](https://github.com/jattasNI/typescript-eslint-sandbox/blob/main/src/index.ts) with all the cases I could think of if you want have a visual aid to think it through.

File: typescript-requiring-type-checking.js (hunk 10 above)
Comment (rajsite) [reply typescript-requiring-type-checking.js:27]: Tried to think of what one-liner patterns we would use to implement nullishOrEmpty if this rule is enabled and could only think of the following:
```ts
const myFunc = (input: string | null): void => {
    const coerceNull = input ?? 'default';

    // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
    const coerceNullOrEmpty1 = input || 'default';

    const coerceNullOrEmpty2 = (input === '' ? undefined : input) ?? 'default';

    const coerceNullOrEmpty3 = input === null || input === undefined || input === '' ? 'default' : input;

    const coerceNullOrEmpty4 = typeof input === 'string' && input !== '' ? input : 'default';
};
```
If we were not in a TS world then I'd lean to option 4 since it is very explicit about what is allowed; option 4 has the strongest guarantee that the result is a string.

Since we are in a TS world then option 4 is not necessary and the clearest representation of "nullishOrEmpty" may be the logical OR. Since we are in a TS world and assuming codebases do not proliferate `any` types, etc. then I guess it makes sense to allow logical OR coalescing...

typescript-requiring-type-checking.js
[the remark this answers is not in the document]
rajsite: I think the following is the configuration we want to check (allowing strings but not numbers in boolean expressions):
```js
        '@typescript-eslint/strict-boolean-expressions': ['error', {
            allowString: true,
            allowNumber: false,
            allowNullableString: true,
            allowNullableNumber: false,
            allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing: true,
        }],
```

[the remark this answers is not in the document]
TrevorKarjanis: In testing, this rule does not work as expected. I think my suggestion to test was misguided and later corrected by Jesse. The `allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing` option does in fact error to report that strictNullChecks is disabled. In the documentation it states, "If for some reason you cannot turn on strictNullChecks, but still want to use this rule - you can use this option to allow it - but know that the behavior of this rule is undefined with the compiler option turned off." This rule requires strictNullChecks and will report, "Unexpected object value in conditional. The condition is always true." on any object without it, even when combined with an operator, e.g. `if (object && object.length)`.

**Proposal**
Leave this rule disabled, but provide a comment recommending its usage with strictNullChecks. The following is slightly adjusted from Milan's based on default values.
```javascript
/*
    This rule requires the TypeScript strictNullChecks compiler flag. If strictNullChecks is enabled, disallow numbers 
    with the following configuration.
    '@typescript-eslint/strict-boolean-expressions': ['error', {
        allowNumber: false,
        allowNullableBoolean: true,
        allowNullableString: true,
        allowNullableNumber: false
    }]
*/
'@typescript-eslint/strict-boolean-expressions': 'off'
```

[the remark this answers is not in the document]
mure: What is the point then of allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing if it still errors on having strictNullChecks disabled?

typescript-requiring-type-checking.js
mure: **Updated proposal**
```js
/* 
    The default configuration for this rule requires that enum members
    follow the camelCase style, which doesn't match the existing convention
    within NI and the broader TS community to use PascalCase. The rest of the
    configuration is identitical to the default.
*/ 
"@typescript-eslint/naming-convention": [
    "error",
    {
        selector: "enumMember",
        format: ["PascalCase"],
    },
    {
        selector: "default",
        format: ["camelCase"],
        leadingUnderscore: "allow",
        trailingUnderscore: "allow",
    },
    {
        selector: "variable",
        format: ["camelCase", "UPPER_CASE"],
        leadingUnderscore: "allow",
        trailingUnderscore: "allow",
    },
    {
        selector: "typeLike",
        format: ["PascalCase"],
    },
],
```

It doesn't appear that it's possible to extend the default configuration, so we will have to duplicate it.

typescript-requiring-type-checking.js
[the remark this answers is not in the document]
TrevorKarjanis: When false, it ensures the strictNullChecks compiler option is enabled. When true, it allows the rule to be run without strictNullChecks, but its behavior is undefined. I think it ended up in this state, because the author originally expected all users to have strictNullChecks enabled by default if they were using this rule. That seems like a ridiculous assumption given the history of most languages and the compiler's default values, but in the end this rule and some others didn't land in the recommended set for that reason. The authors would prefer to simply [silence the rule](https://github.com/typescript-eslint/typescript-eslint/pull/2345#issuecomment-667384255) if it were ever included in the recommended set.

typescript-requiring-type-checking.js
rajsite: I voted on Jesse's but have an alternate proposal comment that is stricter and aligns with Trevor's strictNullChecks recommendation:
```
/*
    Both nullish coalescing (`??`) and logical OR (`||`) may be used 
    to provide default values in specific cases.

    Developers should choose the strictest operator for their use case:
    - `??` to provide a default for nullish values (null or undefined) 
    - `||` to provide a default for values that are nullish or one of: empty string / boolean false.

    Logical OR checks should not be used for other falsy values such as numbers (0, -0, 0n, NaN).
 */
'@typescript-eslint/prefer-nullish-coalescing': 'off',
```

typescript-requiring-type-checking.js
jattasNI: @rajsite this thread is ready to close

typescript-requiring-type-checking.js
jattasNI: @rajsite this thread is ready to close. @LukasKall hasn't voted on the latest proposal but it's aligned with what we've talked about before so it's likely he's ok with it. If not, we can revisit.

typescript-requiring-type-checking.js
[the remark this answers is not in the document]
jattasNI: @rajsite this thread is ready to close. @LukasKall hasn't voted on the latest proposal but it's aligned with what we've talked about before so it's likely he's ok with it. If not, we can revisit.

typescript-requiring-type-checking.js
jattasNI: @rajsite this thread is ready to close using your revised comment above. @LukasKall voted opposite to where we landed in person. If Milan's argument isn't convincing enough for him to change his vote, we can revisit.

LukasKall: I'm fine with Milan's proposal so this thread can be closed.

rajsite:
This PR was closed in favor of #37

PR #21, opened 2021-03-26
Add recommended IDE configuration to README.md

jattasNI:
@rajsite suggested adding info to the readme about how to turn on live linting feedback in popular IDEs.

PR #6, opened 2020-10-20
Update to single package repo and add GitHub Action

rajsite:
- Moves the `@ni/eslint-config` package to the root of the repository
- Adds pull request and branch build GitHub Action workflows
- Adds release GitHub Action workflow for tagged commits

PR #22, opened 2021-03-27
Update repo logo to include TypeScript logo

jattasNI:
Given NI's preference for TypeScript code, it seemed wrong to only show a JS logo.

I modified the logo using an image editor to include the official TS logo too.

Also updated the README to load the logo via a relative path so you can see logo changes when in a branch.


resources/logo.svg
```diff
@@ -1,46 +1,44 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8" standalone="yes"?>
```
TrevorKarjanis, line 1: I would suggest keeping the logos in separate files. This leaves them more easily maintainable, especially given that one is from the design team who also has a copy.

README.md
```diff
@@ -48,4 +48,4 @@ Extend @ni/eslint-config/typescript in the ESLint configuration. Configure the @
 
 ## License
 
-[MIT (c) 2020 National Instruments Corporation](./LICENSE)
\ No newline at end of file
+[MIT (c) 2020 National Instruments Corporation](./LICENSE)
```
TrevorKarjanis, line 51: Should we update this to 2021?

README.md
```diff
@@ -1,6 +1,6 @@
 # NI JavaScript and TypeScript Style Guide
 
-![Logo](https://raw.githubusercontent.com/ni/javascript-styleguide/HEAD/resources/logo.svg)
+![Logo](./resources/logo.svg)
```
rajsite, line 3: The githubusercontent path is so the logo is not broken on pages like the npm published page: https://www.npmjs.com/package/@ni/eslint-config

See: https://github.com/ni/javascript-styleguide/pull/6/files#r508710719

Can you restore the path and add a comment to the README ie
```
<!-- Preserve the raw.githubusercontent.com path to enable image rendering when viewed from https://www.npmjs.com/package/@ni/eslint-config -->
```

rajsite (changes requested):
Some additional discussion on the teams channel

README.md
jattasNI: Good suggestion. I ended up doing this in #21

README.md
jattasNI: Good suggestion. I ended up doing this in #21

jattasNI:
Given the feedback in this PR and on our Teams chat, I'm going to abandon this PR and submit a request for a new logo from our design team.

PR #23, opened 2021-03-27
Add no-restricted-globals to block fit and fdescribe

jattasNI:
The Jasmine 'fdescribe' and 'fit' functions are handy for local development but should not be committed to production.

Similar to [the WebVI codebase in the ASW repo](https://dev.azure.com/ni/DevCentral/_git/ASW?path=%2F.eslintrc.json&version=GBmain&line=22&lineEnd=23&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents), we should block them via no-restricted-globals.

A possible alternate approach would be to use [eslint-jasmin-plugin](https://github.com/tlvince/eslint-plugin-jasmine/blob/master/docs/rules/no-focused-tests.md) but this approach is simpler.


LukasKall (approved):

TrevorKarjanis (approved):

rajsite (approved):

rajsite:
Relatively small change I don't think is controversial, @mure if you have any feedback we can do a follow-up pr.

PR #25, opened 2021-04-02
Disable the max-len rule.

TrevorKarjanis:
- Disable the max-len rule.

Tested


jattasNI (approved):

rajsite (approved):

LukasKall (approved):

rajsite:
Relatively small change I don't think is controversial, @mure if you have any feedback we can do a follow-up pr.

PR #27, opened 2021-04-06
Disable the prefer-destructuring rule.

TrevorKarjanis:
- Disable the [prefer-destructuring](https://eslint.org/docs/rules/prefer-destructuring) rule.

Notes
- I tested this rule following previous concern about its necessity and found 73 errors in one codebase most of which were a complete waste of the developers' time. There are many single variable use cases where destructuring is actually not preferred.
- Examples:
```typescript
const ruleId = this.route.snapshot.queryParams.ruleId;
// Even in this multi-variable case, destructuring doesn't pose enough of an advantage.
this._startDate = this.range[0];
this._endDate = this.range[1];
// Are you kidding me?
const trigger = this.trigger;
// Must be rocket science.
const featureId = DataAlarmRuleHelper.getFeatureConfig(this.rule).featureId;
```

Tested
- Verified that the errors are removed

### DIFF
--- a/index.js
+++ b/index.js
@@ -213,6 +213,12 @@ module.exports = {
         'object-shorthand': ['error', 'properties', { avoidQuotes: true }],
 
         /*
+            Destructuring should be considered for multi-variable declarations and assignments while simple single
+            variable declarations and assignments likely don't require it.
+        */
+        'prefer-destructuring': 'off',
+
+        /*
             This configuration already supports the JSDoc syntax. Add additional syntax as line or
             block exceptions or markers when necessary.
         */


jattasNI (approved):

index.js
```diff
@@ -212,6 +212,12 @@ module.exports = {
         */
         'object-shorthand': ['error', 'properties', { avoidQuotes: true }],
 
+        /*
+            Destructuring should be considered for multi-variable declarations and assignments while simple single
+            variable declarations and assignments likely don't require it.
+        */
+        'prefer-destructuring': 'off',
```
rajsite, line 219: I don't see a built-in rule to enforce it but I'd go a step further and say in the comment we should discourage [destructuring with renaming](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assigning_to_new_variable_names).

rajsite (approved):

LukasKall (approved):
I like to use the rule when accessing multiple properties of an object. That's also the rationale in the Airbnb styleguide for enabling this rule. It's the same for arrays, I just think it looks cleaner. The [Airbnb styleguide](https://github.com/airbnb/javascript#destructuring) has some more rationale for it.

```javascript
const { firstName, lastName } = user;

[this._startDate, this._endDate] = this.range;
```

I agree that it doesn't give much value when accessing only a single property/element and since I don't see a built-in configuration to enforce the rule only for multiple properties/elements, I'm fine with disabling it and relaying on code reviews.

index.js
mure: Why do you think we shouldn't allow this?

mure (approved):

index.js
TrevorKarjanis: Renaming is fine and often required.

rajsite: I find the syntax of destructuring with renaming to not be that much less verbose but I do find it less clear (I can scan the second one and understand the outcome quicker). If I ask myself, what are the properties of o that this depends on, it's more readily apparent in the second example.

```js
const {a: awesomeA, b: awesomeB} = o;
```
vs
```js
const awesomeA = o.a;
const awesomeB = o.b;
```

I find visually parsing it more difficult than without destructuring, particularly with destructuring in function parameters:
```js
function test (something, {a: awesomeA, b: awesomeB} = config) {...}
```
vs
```js
function test (something, config) {
    const awesomeA = config.a;
    const awesomeB = config.b;
}
```

Particularly if you add defaults and have more than two properties it becomes a visual soup of brackets and things that look like object literals. The syntax for destructuring with renaming is so much like the object literal syntax that I find it a code scanning speedbump. It's the same reason I advocated for TypeScript interface definition over the type alias format (where type aliases look a lot like object literals).

But I also don't feel strongly about it. I'll align with the majority vote. Do @jattasNI or @LukasKall have an opinion?

LukasKall: I agree that at a certain point, destructuring with renaming becomes very complex and less clear. But I don't feel like we should discourage destructuring with renaming at all. It's a personal choice at which point it's to complex and in some cases its quite useful, e.g. when destructuring multiple properties and only renaming one.

```javascript
const { a, b, c: betterC } = obj;
```

jattasNI: I haven't seen destructuring used much since it's a new-ish feature so all of the syntaxes are harder to read for me than splitting it into separate assignments. But the more I see the syntax (including the rename version), the more I like it so I don't think we should discourage it.

rajsite: sounds good to me! thanks for the discussions 👍

rajsite (approved):

PR #28, opened 2021-04-06
Disable the no-continue rule.

TrevorKarjanis:
- Disable the [no-continue](https://eslint.org/docs/rules/no-continue) rule.

Notes
- This rule produced 39 errors.
- It doesn't look like we addressed this rule in the [airbnb review](https://ni.visualstudio.com/Users/_git/abarreto-misc/pullrequest/83582?_a=files&path=%2Fairbnb-js-review%2Frules%2Fstyle.js) line 279.
- Continue statements are a common programming construct, and our developers are capable of using it properly.
- The examples in the rule's documentation are questionable.

Tested
- Verified that the errors are removed

### DIFF
--- a/index.js
+++ b/index.js
@@ -97,7 +97,8 @@ module.exports = {
         'no-console': 'error',
 
         /*
-            Continue statements are a common, useful, and can improve readability. Use of labels is disallowed.
+            `continue` statements can improve readability but care should be taken to avoid confusing control flow.
+            They should generally be the only statement within an `if` block and should never use labels.
         */
         'no-continue': 'off',
 
--- a/index.js
+++ b/index.js
@@ -97,6 +97,11 @@ module.exports = {
         'no-console': 'error',
 
         /*
+            Continue statements are a common, useful, and can improve readability. Use of labels is disallowed.
+        */
+        'no-continue': 'off',
+
+        /*
             There may be exceptions, like array iterators, where it is appropriate to disable this
             rule with an inline comment.
         */


jattasNI (approved):

rajsite:
I'm not in-favor of disabling this rule. I think it makes sense to discourage the use because it is a relatively under used feature (you found relatively few instances) and anecdotally I find code that is refactored to not use continue statements easier to read or equally readable.

I think there are cases, like streaming parsing where the algorithm lends naturally to using continue or usage in hot loops but I think it should be generally discouraged. I agree that developers can generally use it properly but I think it adds cognitive overhead in review and refactor because it's used infrequently and I would generally discourage use.

index.js
```diff
@@ -96,6 +96,11 @@ module.exports = {
         */
         'no-console': 'error',
 
+        /*
```
rajsite, line 99: I'm not in-favor of disabling this rule. I think it makes sense to discourage the use because it is a relatively under used feature (you found relatively few instances) and anecdotally I find code that is refactored to not use continue statements easier to read or equally readable.

I think there are cases, like streaming parsing where the algorithm lends naturally to using continue or usage in hot loops but I think it should be generally discouraged. I agree that developers can generally use it properly but I think it adds cognitive overhead in review and refactor because it's used infrequently and I would generally discourage use.

rajsite (changes requested):

mure:
Should we vote on this like we did other rules?

index.js
TrevorKarjanis: This is a commonly used feature across many languages. It's very readable, easily learned, and requires minimal cognitive overhead.

jattasNI: I also find myself having to think a little more when I see `continue` in code and generally prefer to refactor to avoid it. But I lean towards allowing it since:
1. our C# conventions don't disallow it
2. there are cases where it's perfectly understandable and even preferable, like when used at the start of a block to avoid extra nesting
```ts
// with continue at the beginning of a block: easy to understand and avoids 1 level of nesting in the main section of code
for (const val in array) {
    if (shouldSkip(val)) {
       continue;
    }
    // a bunch more code that has its own nesting
}

// without continue, extra nesting for the main section of code
for (const val in array) {
    if (!shouldSkip(val)) {
        // a bunch more code that has its own nesting
    } 
}
```

*Proposal:* allow `continue` but update the comment to discourage poor usage
```
/* `continue` statements can improve readability but care should be taken to avoid confusing control flow. They should generally be the only statement within an `if` block and should never use labels. */
'no-continue': 'off',
```

jattasNI:
> Should we vote on this like we did other rules?

For single-rule PRs like this one, let's use the review approval as our overall voting mechanism. But if it generates discussion like this one did, then we can go back to our comment 👍 voting system on various proposals.

jattasNI (approved):

index.js
jattasNI: We agreed in person to go with this proposal. @TrevorKarjanis can update with the proposed comment and complete the PR

rajsite: Updated wit the proposed comment 👍

rajsite (approved):

PR #30, opened 2021-04-21
 Elevate recommended TypeScript rules from warning to error

rajsite:
The existing recommended rules used by the typescript.js config have [several rules configured as `warn`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended.ts#L17) instead of as `error`.

This change updates the rules to always error. Interested if exisiting code bases testing the TypeScript ESlint rules noticed the warnings and if we have concerns with making these recommended rules stricter.

The change also updates the `print-typescript-properties.js` tool to make it easier to audit recommended rules.

### DIFF
--- a/typescript.js
+++ b/typescript.js
@@ -19,7 +19,11 @@ module.exports = {
             https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended.ts
         */
 
-        // None
+        // The following elevate recommended rules from warning to error
+        '@typescript-eslint/explicit-module-boundary-types': 'error',
+        '@typescript-eslint/no-explicit-any': 'error',
+        '@typescript-eslint/no-non-null-assertion': 'error',
+        '@typescript-eslint/no-unused-vars': 'error',
 
         /*
             TypeScript rules outside of the recommended configuration
--- a/tools/print-typescript-properties.js
+++ b/tools/print-typescript-properties.js
@@ -28,6 +28,21 @@ const typeScriptRequiringTypeChecks = Object.keys(plugin.rules)
     .filter(key => !recommended(key))
     .filter(key => requiresTypeChecking(key));
 
+const typeScriptRecommended = Object.keys(plugin.rules)
+    .filter(key => !extendsBaseRule(key))
+    .filter(key => recommended(key))
+    .filter(key => !requiresTypeChecking(key));
+
+const typeScriptRecommendedTypeChecks = Object.keys(plugin.rules)
+    .filter(key => !extendsBaseRule(key))
+    .filter(key => recommended(key))
+    .filter(key => requiresTypeChecking(key));
+
+global.console.log('Note: When printing this does not print the default config and assumes "error"');
+global.console.log('TypeScript Recommended:');
+print(typeScriptRecommended);
+global.console.log('TypeScript Recommended Requiring Type Checks:');
+print(typeScriptRecommendedTypeChecks);
 global.console.log('TypeScript Extensions:');
 print(typeScriptExtensions);
 global.console.log('TypeScript Extensions Requiring Type Checks:');


LukasKall (approved):

typescript.js
```diff
@@ -19,7 +19,11 @@ module.exports = {
             https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended.ts
         */
 
-        // None
+        // The following elevate recommended rules from warning to error
+        '@typescript-eslint/explicit-module-boundary-types': 'error',
```
TrevorKarjanis, line 23: [@typescript-eslint/explicit-module-boundary-types](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md)
We already enabled [@typescript-eslint/explicit-function-return-type](https://github.com/ni/javascript-styleguide/blob/2aa42157a52b67ec25615ee9f0f370ea01afa768/typescript.js#L53).
**Proposal**
Assuming we're extending the recommended set (are we?), disable the rule.
`'@typescript-eslint/explicit-module-boundary-types': 'off'`

typescript.js
```diff
@@ -19,7 +19,11 @@ module.exports = {
...
+        '@typescript-eslint/no-explicit-any': 'error',
+        '@typescript-eslint/no-non-null-assertion': 'error',
```
TrevorKarjanis, line 25: [@typescript-eslint/no-non-null-assertion](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-non-null-assertion.md)
We haven't enabled strict null checking, but non-null assertions are the only reason I would consider it. They enable exceptions in testing where a null check would otherwise not fix/hide a bug and prevent invalid code paths. They assert [non-null and non-undefined](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-and-non-undefined-type-guards).

typescript.js
```diff
@@ -19,7 +19,11 @@ module.exports = {
...
+        '@typescript-eslint/no-unused-vars': 'error',
```
TrevorKarjanis, line 26: [@typescript-eslint/no-unused-vars](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md)
This is an extension rule. It should be grouped with extensions, and the ESLint equivalent should be disabled.

TrevorKarjanis (changes requested):

typescript.js
LukasKall: Do we have a rule the enforces explicit types for function arguments or is it just enforced by the TypeScript compiler option `noImplicitAny`? I think @typescript-eslint/explicit-module-boundary-types enforces this, but @typescript-eslint/explicit-function-return-type does only enforce the return type to be explicit.

TrevorKarjanis: Good catch, I haven't found one. That brings up a few additional points.
- I do think `noImplicitAny` would enforce the parameters equivalently. This rule allows default parameters.
- If we aren't enabling `noImplicitAny`, do we really need to document every module as if it were public, like a library? Most application modules are not public. I'm not against enabling `noImplicitAny`, though. If we do enable it, the less rules the better performance. ESLint is quite slow in testing.

typescript.js
rajsite: @TrevorKarjanis could you give a specific proposal? In discussion it sounded like we would want to disable the non-null assertion operator with or without strict null checking. Do you have examples where it is useful to enable?

TrevorKarjanis: If you don't have strict null checks enabled, you're asserting non-null anytime you don't perform a null check, so we actually do this all the time. It's a good thing. In cases where null is unexpected, it's best to allow an exception to be thrown, so that the bug can be fixed. In most cases, the application would be unusable anyway. In the case of HTTP, Angular handles this well by enforcing a value exist before initializing a view with resolvers. The same argument can be used for strict initialization. However, TypeScript provides a valuable operator to assert that a value is expected to be non-null. With strict null checks enabled, this allows the developer to state that null is unexpected, a code path to handle it doesn't make any sense (the app is unusable or, even better, shouldn't have been routed to), and an exception should be thrown if it's encountered. Success handlers for HTTP requests are a common example, but we do this in other sub-routines all the time where null isn't possible or is checked for early on.

When I tested strict null checks, I received 768 errors, and that's one-third of the entire application. We simply couldn't use strict null checks without this operator, and that makes complete sense. If you choose to enable strict null checks, you have to design your application around it, and you can use this operator to do that.

TLDR; This rule requires strict null checking, and as far as I can tell the null assertion operator is harmless if it's disabled. Non-null assertions are expected to be used with strict null checking and are the reason strict null checks are being adopted in the Angular community. This rule either applies to certain application types or is naive.

TrevorKarjanis: [Edited after Testing] If you don't have strict null checks enabled, you're asserting non-null anytime you don't perform a null check, so we actually do this all the time. It's a good thing. In cases where null is unexpected, it's best to allow an exception to be thrown, so that the bug can be fixed. In most cases, the application would be unusable anyway. In the case of HTTP, Angular handles this well by enforcing a value exist before initializing a view with resolvers. The same argument can be used for strict initialization. However, TypeScript provides a valuable operator to assert that a value is expected to be non-null. With strict null checks enabled, this allows the developer to state that null is unexpected, a code path to handle it doesn't make any sense (the app is unusable or, even better, shouldn't have been routed to), and an exception should be thrown if it's encountered. HTTP request handlers don't require this, because null is not included in the type for the response (hence, we're already asserting non-null). However, it's very useful for making assertions when a code path and business logic dictate an expected value. A common example is return values from browser and third-party APIs, e.g. Array.find, where the original creator can't make assumptions. Most commonly it's used to make assertions about uninitialized class properties like components that are populated with data from a resolver that guarantees a value. Lastly, the operator is also used in testing to allow errors to propagate.

When I tested strict null checks, I received 768 errors, and that's one-third of the entire application. We simply couldn't use strict null checks without this operator, and that makes complete sense. If you choose to enable strict null checks, you have to design your application around it, and you can use this operator to do that.

TLDR; This rule requires strict null checking, and as far as I can tell the null assertion operator is harmless if it's disabled. Non-null assertions are expected to be used with strict null checking in application and test code. This rule either applies to certain application types or is naive. I would only expect this rule to be enabled if the operator were being abused—which it's not and wouldn't make sense to.

jattasNI: @TrevorKarjanis would this be an accurate summary of your proposal?

/* If an application uses the `strictNullChecks` tsconfig option then non-null assertions are a useful escape hatch.
    If an application doesn't use the `strictNullChecks` tsconfig option then non-null assertions are harmless. */
        '@typescript-eslint/no-non-null-assertion': 'off',

TrevorKarjanis: Sorry, I should have proposed a comment. I would just reword escape hatch so that it doesn't like a workaround, "If an application uses the `strictNullChecks` tsconfig option then non-null assertions are a useful for asserting when null values are not expected."

rajsite: If I understand the discussion correctly the non-null assertion operator is only useful for codebases using strictNullChecks (I'm still on the fence about why it's useful to provide an assertion override for a Type instead of correcting the type when possible but in cases you don't own the type and don't want to create an override I can understand).

However I don't see value in using the operator in code bases without strictNullChecks. It is a no-op type-wise since null checks are not enforced and also a [no-op in generated code](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator). It just adds visual noise and confusion in [some cases](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#breaking-changes). Why would we allow usage in codebases with strictNullChecks disabled (which I expect to be most repos for the forseeable future)?

Proposal:
```
/*
    For applications that do not use `strictNullChecks` , non-null assertions do not provide additional type safety and if used inconsistently add to visual noise so are discouraged.
    If an application uses the `strictNullChecks` tsconfig option then non-null assertions are a useful escape hatch and applications may allow usage in their local eslint configuration.
*/
        '@typescript-eslint/no-non-null-assertion': 'error'
```

typescript.js
rajsite: Good catch! Thanks!

typescript.js
TrevorKarjanis: We would allow the operator, because it should absolutely be available if strict null checks are enabled. Visual noise is not an issue and no one is using the operator anyway. The operator's purpose is to provide corrections to the compiler.

typescript.js
jattasNI: Agreed in person to leave this rule enabled as an error.

typescript.js
jattasNI: Consensus that we want this to be 
- `error` for codebases with `strictNullChecks` off 
- `off` for codebases with `strictNullChecks` on. 

Since most codebases have `strictNullChecks` off, we agreed in person to set this to `error` but provide documentation explaining when it's ok to disable. This is basically rajsite's proposal, but change the language away from "escape hatch". @TrevorKarjanis to make an alternate proposal for the language in the comment.

TrevorKarjanis: In short, the type system isn't all knowing. A type that originally includes null can loose null based on the code path and input values as defined by the business logic. Therefore, the non-null assertion operator is useful when strictNullChecks are enabled. Nor did we decide to forbid type casting. The cast in this case is to assert that the value can no longer be null.
**Proposal**
```javascript
/*
    With `stickNullChecks disabled, non-null assertions do not provide additional type safety, and if used inconsistently,
    add to visual noise so are discouraged. With `strictNullChecks` enabled, the non-null operator is useful for asserting
    that a value no longer be null, and their usage should be allowed.
*/
'@typescript-eslint/no-non-null-assertion': 'error'
```

PR #33, opened 2021-05-07
Add print-evaluated-rules script

rajsite:
The `print-evaluated-rules` script is intended to be used during development to make it easier to audit different configuration settings.

The `print-evaluated-rules` script will use the eslint api to to run the lint configuration evaluation on the different smoke test files. The smoke test files should represent all of the available lint configurations. The calculated configs show the final evaluated rules for a file and make it easier to audit some behavior instead of cross-referencing all the rulesets individually (airbnb, typescript recommended, etc.).

Note: This script does help identify some rules set to warn such as [no-constant-condition](https://github.com/ni/javascript-styleguide/pull/33/checks?check_run_id=2531590851#step:10:1170) which will be addressed in a follow-up PR.

PR #34, opened 2021-05-12
Elevate remaining warn rules and enable rules from WebPortal

rajsite:
This PR enables rules from two investigations:
1. The PR #33 added a script to evaluate all rules that evaluate to a "warn" state. This PR cleans up the remaining rules that evaluate to "warn" and changes them to error.
2. When WebPortal was migrated from tslint to eslint using the ni eslint rules there were [several rules not covered by this ruleset](https://ni.visualstudio.com/DevCentral/_git/op-web-portal/pullrequest/158284). It turns out all of those rules will be enabled in future versions of the airbnb ruleset. This change enables those rules preemptively to avoid loss of rule coverage migrating from tslint to eslint.

### DIFF
--- a/index.js
+++ b/index.js
@@ -116,14 +116,14 @@ module.exports = {
 
         /*
             Disallow duplicate conditions in if-else-if chain.
-            This rule does not need to be explicilty set after enabled in airbnb:
+            This rule does not need to be explicitly set after enabled in airbnb:
             https://github.com/airbnb/javascript/blob/1eadb93e377da1e56c3f91f26610e5d0a00738a9/packages/eslint-config-airbnb-base/rules/errors.js#L44
         */
         'no-dupe-else-if': 'error',
 
         /*
             Disallow assigning to imported bindings.
-            This rule does not need to be explicilty set after enabled in airbnb:
+            This rule does not need to be explicitly set after enabled in airbnb:
             https://github.com/airbnb/javascript/blob/5641278fa1df75ba3b6af2959a553bd682c0ad42/packages/eslint-config-airbnb-base/rules/errors.js#L83
         */
         'no-import-assign': 'error',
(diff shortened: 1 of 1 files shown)


alejandro5042 (approved):

LukasKall (approved):

jattasNI (approved):

index.js
```diff
@@ -102,6 +108,26 @@ module.exports = {
...
 
+        /*
+            Constant expressions in conditions are likely an error. Cases representing an infinite loop
+            may utilize the construct `for (;;) {}`.
+        */
+        'no-constant-condition': 'error',
+
+        /*
+            Disallow duplicate conditions in if-else-if chain.
+            This rule does not need to be explicilty set after enabled in airbnb:
```
TrevorKarjanis, line 119: Misspelled "explicitly".

index.js
```diff
@@ -179,6 +205,13 @@ module.exports = {
                 message: '`with` is disallowed in strict mode because it makes code impossible to predict and optimize.',
             }],
 
+        /*
+            Disallow returning values from setters.
+            This rule does not need to be explicilty set after enabled in airbnb:
+            https://github.com/airbnb/javascript/blob/5641278fa1df75ba3b6af2959a553bd682c0ad42/packages/eslint-config-airbnb-base/rules/errors.js#L121
+        */
+        'no-setter-return': 'error',
```
TrevorKarjanis, line 213: This can be disabled for TypeScript, because it is covered by compiler error [2408](https://github.com/microsoft/TypeScript/blob/dc4cd52a9a09bd4eee1fbebde98b41f2c503b335/src/compiler/diagnosticMessages.json#L1242).

TrevorKarjanis (approved):

index.js
rajsite: This seems to be handled internally somehow. With the rule enabled in the index.js settings and doing the computed ruleset on a ts file I can see the rule resolves to being enabled. However when eslint runs on a ts file and in the editor it ignores the eslint rule (the eslint rule error for no-setter-return does not error but other eslint rules do):

image

So I think typescript-eslint is taking care of that for us in some way.

TrevorKarjanis: Interesting, maybe another rule that checks the configuration? No biggie, if it works it works.

index.js
rajsite: 🤦‍♂️

mure (approved):

PR #37, opened 2021-05-19
Enable TypeScript Rules Requiring Type Checking (actual)

rajsite:
The goal of this PR is to evaluate rules used in `typescript-requiring-type-checking.js`.

A previous PR (#20) started the process of enabling the rules requiring type checking but the branch of the PR needed to be recreated. See the previous PR for thorough discussion of rule configuration. This PR includes some additional README tweaks and wording updates.


README.md
```diff
@@ -34,22 +37,31 @@ See [instructions below](#usage-angular) to use a schematic to install dependenc
...
-```json
+```js
+{
+    extends: '@ni/eslint-config/typescript'
+}
+```
+
+### TypeScript Rules Requiring Type Checking
+
+Extend `@ni/eslint-config/typescript-requiring-type-checking` in a separate ESLint configuration file (such as `.eslintrc-requiring-type-checking.js`) that includes a TypeScript configuration for `parserOptions.project`. Due to the longer execution time, the configuration requiring type checking should only be run from a `npm run lint` command explicitly invoked by developers and on CIs.
```
jattasNI, line 58: Suggestion to distinguish the different lint commands a bit more clearly:

"...should not be included in the main `npm run lint` command. Prefer a separate `npm run lint-types` command explicitly invoked by developers and on CIs."

typescript-requiring-type-checking.js
```diff
@@ -10,28 +10,121 @@ module.exports = {
...
+
+            Logical OR checks should not be used for other falsy values such as numbers (0, -0, 0n, NaN).
+        */
+        '@typescript-eslint/prefer-nullish-coalescing': 'off',
+
+        '@typescript-eslint/prefer-readonly': 'error',
+
+        /*
+            This rule requires adding `readonly` keywords to all levels of non-primitive parameters in order to prevent
+             any part of the parameter from being mutated. While the spirit of preferring not to mutate inputs is correct,
```
jattasNI, line 100: Extra indentation space before " any part"

jattasNI (approved):

jattasNI:
Doesn't have to be in this PR, but reminder of [this thread](https://github.com/ni/javascript-styleguide/pull/20/files#r609120524) where we said we'll need to upgrade our typescript-eslint version at some point to update a bug fix to one of the fixers.

README.md
rajsite: Oh interesting. I was thinking there would be a single `npm run lint` command that runs both configurations. I expect during development most folks are relying on IDE feedback. I was thinking the cli would be run infrequently, ie when I'm ready to push and not continuously during development.

rajsite:
> upgrade our typescript-eslint version at some point to update a bug fix to one of the fixers

@jattasNI  I think that's a good follow-up PR. I'd gate the 1.0 release on that.

README.md
TrevorKarjanis: How does this ensure that these rules are considered by IDE plugins. From the [documentation](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/TYPED_LINTING.md#performance):

"Most of our users are fine with this, as they think the power of type-aware static analysis is worth it. Additionally, most users primarily consume lint errors via IDE plugins which, through some caching magic, do not suffer the same penalties."

index.js
```diff
@@ -26,6 +26,14 @@ module.exports = {
         */
         'comma-dangle': ['error', 'only-multiline'],
 
+        /*
+            Always provide a `default` case in `switch` statements.
+            If the default case is logically unreachable, throw an Error.
+            If the default case should do nothing, add a comment explaining why
+            to indicate that it is an intentional decision.
+        */
+        // Notes for the Airbnb 'default-case' rule configuration
```
TrevorKarjanis, line 35: What's the point of this comment and the one for `no-floating-promises` below?

typescript-requiring-type-checking.js
```diff
@@ -10,28 +10,121 @@ module.exports = {
...
 
-        // None
+        /*
+            The jasminewd2 library used by Angular applications will result in a significant number of floating promises.
+            Consider disabling the '@typescript-eslint/no-floating-promises' rule for tests that utilize jasmine expects.
+        */
+        // Notes for typescript-eslint recommended '@typescript-eslint/no-floating-promises' configuration
+
+        /*
+            Allow usage of String.prototype.match for regular expression matching
```
TrevorKarjanis, line 20: Terminate with a period.

TrevorKarjanis (approved):

README.md
jattasNI: Ah, good points. I was confused on multiple fronts but now that I understand how it works a little better, I agree that we don't need a separate `npm run lint-types` command.

If that "caching magic" is quick enough, then maybe we don't actually need a separate config file either. Let's discuss in our meeting today.

jattasNI: We agreed in person to make our default recommendation to include type checking in IDEs and on builds. We can add a note mentioning splitting them as an option if performance is bad.

index.js
rajsite: I merged the comment showing the rule name and the comment with the rule note together for these two cases.

jattasNI (approved):

README.md
```diff
@@ -1,60 +1,67 @@
...
+Alternatively, use [`npx install-peerdeps`](https://www.npmjs.com/package/install-peerdeps) as a shortcut to install the packages for you.
 
 ```bash
 npx install-peerdeps --dev @ni/eslint-config
 ```
 
-### Angular
+### Additional Angular-specific Installation
 
-See [instructions below](#usage-angular) to use a schematic to install dependencies.
```
TrevorKarjanis, line 29: I chose this anchor name, "usage-angular", specifically to avoid other Angular subsections in the future as demonstrated in another comment.

README.md
```diff
@@ -1,60 +1,67 @@
...
 
-Alternatively, use [install-peerdeps](https://www.npmjs.com/package/install-peerdeps) as a shortcut to install the packages for you.
+Alternatively, use [`npx install-peerdeps`](https://www.npmjs.com/package/install-peerdeps) as a shortcut to install the packages for you.
 
 ```bash
 npx install-peerdeps --dev @ni/eslint-config
 ```
 
-### Angular
+### Additional Angular-specific Installation
```
TrevorKarjanis, line 30: There's no need for "-specific". "Angular Installation" is fine and simple if "Installation" is needed at all. I prefer less words in the instructions section of a README, because most users don't care and only want exactly what they need to do and fast.

README.md
```diff
@@ -1,60 +1,67 @@
...
 }
 ```
 
 ### TypeScript
 
-Extend `@ni/eslint-config/typescript` in the ESLint configuration. Configure the `@typescript-eslint` plugin and the project's TypeScript configuration.
+Extend `@ni/eslint-config/typescript` and `@ni/eslint-config/typescript-requiring-type-checking` (type checked rules) in the [ESLint configuration](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats).
 
-```json
+The type checked rules will also require the `parserOptions.project` configuration to be set to the project's TypeScript configuration. The type checked rules run a bit slower as they utilize the TypeScript compiler for type information.
```
TrevorKarjanis, line 50: I think we can move information about rules that require type checking to the performance section below and reduce the word count here. This information doesn't have to be at the forefront as subsequent sections work just as well as reference links. The following combines concise instructions here and performance below. Again, users hate reading and want to flow through the setup fast.

"Extend `@ni/eslint-config/typescript` and `@ni/eslint-config/typescript-requiring-type-checking` in the [ESLint configuration](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats). Set the `parserOptions.project` configuration to the project's TypeScript configuration."

"
## Performance

## TypeScript

`@ni/eslint-config/typescript-requiring-type-checking` includes rules that require type checking that run slower as they utilize the TypeScript compiler for type information. If there are situations where the analysis time for enabling the type checked rules is an excessive burden you may consider creating a separate eslint configuration that avoids extending the type checked rules and omits the `parserOptions.project` configuration to run in specific scenarios. See discussion in the [performance section](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/TYPED_LINTING.md#performance) of the Getting Started - Linting with Type Information guide.

## Angular

Deviations ...
"

TrevorKarjanis (changes requested):

README.md
rajsite: If you look how Github renders it on https://github.com/ni/javascript-styleguide you can see that clicking the first title called angular gives the url with #angular and the second title called angular gives the anchor #angular-1. 

I like to avoid having duplicate titles to avoid usage of the indexed anchors / the need of custom anchors. I went with the option for 'Angular Installation' but if you feel strongly about keeping the indexed anchor + a custom anchor I don't feel super strongly about switching back.

README.md
rajsite: I like it! thanks 👍

README.md
rajsite: responded in other comment

rajsite:
@TrevorKarjanis FYI I reset you in-case there is additional feedback

README.md
TrevorKarjanis: I don't thinks it's a big enough issue in this case. However, the preferred UX should take precedence over such small additional effort unless there's some more serious reason for avoiding indexed anchors.

TrevorKarjanis (approved):

PR #40, opened 2021-05-26
Add linting configurations, tests, and tools for Angular.

TrevorKarjanis:
- Add linting configurations, tests, and tools for Angular.

Notes
- There are two configurations `@ni/eslint-config/angular` for TypeScript and `@ni/eslint-config/angular-template` for HTML.
- Disabling the parse options, which should cause rules that require type checking to fail, didn't result in any errors. I assume then that none of the rules require type checking.

Linted and tested
- Verified that HTML files and inline templates are evaluated
- Verified that the print-angular-properties script works correctly

### DIFF
--- a/angular.js
+++ b/angular.js
@@ -1,0 +1,8 @@
+module.exports = {
+    extends: [
+        'plugin:@angular-eslint/recommended',
+        'plugin:@angular-eslint/recommended--extra',
+        'plugin:@angular-eslint/template/process-inline-templates',
+        './typescript'
+    ]
+};
--- a/package.json
+++ b/package.json
@@ -6,8 +6,10 @@
   "scripts": {
     "prepare": "npm test",
     "test": "eslint .",
+    "test-angular": "eslint ./test/angular/",
     "test-typescript": "eslint ./test/typescript/",
     "test-typescript-typed": "eslint ./test/typescript-type-checking/",
+    "print-angular-properties": "node ./tools/print-angular-properties",
     "print-typescript-properties": "node ./tools/print-typescript-properties",
     "print-evaluated-rules": "node ./tools/print-evaluated-rules"
   },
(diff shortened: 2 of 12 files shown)


angular.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    extends: [
+        'plugin:@angular-eslint/recommended',
+        'plugin:@angular-eslint/recommended--extra',
+        'plugin:@angular-eslint/template/process-inline-templates',
+        './typescript',
+        './typescript-requiring-type-checking'
```
LukasKall, line 7: For TypeScript, we provide two separate rule sets to give clients the option to run the rules requiring type checking separately. Do we want the same for Angular? Or do the angular rules require type checking anyway?

package.json
```diff
@@ -6,8 +6,10 @@
   "scripts": {
     "prepare": "npm test",
     "test": "eslint .",
+    "test-angular": "eslint ./test/angular/",
```
LukasKall, line 9: Do we want to run this in our GitHub pipeline?

LukasKall (commented):
There is nothing in the README about using these Angular configs. Do we want to add that later once we agreed on the rule set?

angular.js
TrevorKarjanis: Actually, it's the other way around. From the description, "Disabling the parse options, which should cause rules that require type checking to fail, didn't result in any errors. I assume then that none of the rules require type checking."

TrevorKarjanis:
> There is nothing in the README about using these Angular configs. Do we want to add that later once we agreed on the rule set?

Correct, I am intentionally waiting until the rule set is defined. For now, users should extend the TypeScript configuration.

package.json
TrevorKarjanis: Yes, I had assumed the pipeline just ran `npm test`, but apparently that is not the case. I will add this to the pipeline.

package.json
```diff
@@ -38,12 +40,25 @@
     "eslint-config-airbnb-base": "~14.2.1"
   },
   "devDependencies": {
```
jattasNI, line 42: Since we now have a significant number of dependencies that won't be needed for non-Angular projects, what do you think of updating the "Installation" section of README.md to provide more explicit guidance about using `npm view` vs `npx install-peerdeps`?

I'm thinking something like a note in the `install-peerdeps` section saying "For non-Angular projects, this will install a significant number of unneeded dependencies"

TrevorKarjanis: @rajsite will verify the behavior of the future version of `npm`. In the spirit of progress, I will move this discussion to an issue.

LukasKall (approved):

jattasNI (approved):

rajsite (approved):

rajsite (changes requested):
Can you update print-evaluated-rules.js to include the angular configuration?

test/angular/.eslintrc.js
```diff
@@ -0,0 +1,17 @@
+// Test TypeScript and templates together to process inline templates.
+module.exports = {
+    overrides: [{
+        extends: '../../angular',
+        files: ['*.ts'],
+        parserOptions: {
+            project: './tsconfig.json',
+            // Resolve the project relative to this file's directory.
+            // https://stackoverflow.com/questions/64933543/parsing-error-cannot-read-file-tsconfig-json-eslint
```
rajsite, line 9: This seems like a useful comment. Can we add it to the other test configuration files for their `tsconfigRootDir` setting?

mure (approved):

TrevorKarjanis:
> Can you update print-evaluated-rules.js to include the angular configuration?

Huh, this must have gotten lost after the branch rename. Yes, I will add it.

TrevorKarjanis:
> @TrevorKarjanis it looks like one of the rules evaluates to "warn", we should make sure to address it in a follow-up PR:

My expectation is to include it in the first evaluation, because we don't expect to use this configuration until at least that is complete.

PR #45, opened 2021-06-16
Override non-recommended rules in the Angular configuration.

TrevorKarjanis:
- Override non-recommended rules in the Angular configuration.

Notes

- All of the non-recommended rules are temporarily defined here so that their configurations can be seen.
- The tests are currently failing, because they violate one of the non-recommended rules.
- To alleviate the difficulty of navigating the source code among many comments, a review containing a comment per rule has already been started.
- Once the evaluation is complete, I will merge our overrides and remove the rest.

Linted and tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,9 @@ configure ESLint for Angular projects** especially when migrating from TSLint. [
     ```bash
     ng g @angular-eslint/schematics:convert-tslint-to-eslint <PROJECT NAME>
     ```
+4. Enable these application-specific rules if your application uses a specific prefix for components or pipes.
+    - [@angular-eslint/component-selector](http://codelyzer.com/rules/component-selector)
+    - [@angular-eslint/pipe-prefix](http://codelyzer.com/rules/pipe-prefix)
 
 ## Usage
 
--- a/angular-template.js
+++ b/angular-template.js
@@ -7,5 +7,51 @@ module.exports = {
             Overrides to Angular template recommended rules:
             https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/src/configs/recommended.json
         */
+
+        /*
+            Overrides to Angular rules outside of the recommended configuration:
+        */
+
+        /*
+            There currently isn't an NI organization wide requirement to enforce accessibility in applications. These
+            rules should be considered if individual applications prioritize it. They may be enabled in the future as
+            the priority of accessibility increases.
+
+            '@angular-eslint/template/accessibility-alt-text': 'error',
+            '@angular-eslint/template/accessibility-elements-content': 'error',
+            '@angular-eslint/template/accessibility-label-has-associated-control': 'error',
+            '@angular-eslint/template/accessibility-table-scope': 'error',
+            '@angular-eslint/template/click-events-have-key-events': 'error',
+            '@angular-eslint/template/mouse-events-have-key-events': 'error',
+            '@angular-eslint/template/no-autofocus': 'error',
+            '@angular-eslint/template/no-positive-tabindex': 'error'
+        */
+
+        '@angular-eslint/template/accessibility-valid-aria': 'error',
+
+        '@angular-eslint/template/conditional-complexity': 'error',
+
+        '@angular-eslint/template/cyclomatic-complexity': 'error',
+
+        /*
+            Enable this rule by default to enforce internationalization for existing applications that are localized
+            and new applications in the chance that they'll need to be localized in the future. Disable this rule
+            if an application will never be localized.
+        */
+        '@angular-eslint/template/i18n': ['error', { checkText: true }],
+
+        '@angular-eslint/template/no-any': 'error',
+
+        /*
+            When considering efficient bindings use memoization and pipes. Avoid heavy array iteratoration, nested
+            function calls, and calls into uncontrolled third-party imports. However, there are many cases where a
+            function call significantly improves template readability and has no impact on performance—where a 1ms
+            execution is imperceivable for users.
+        */
+        '@angular-eslint/template/no-call-expression': 'off',
+
+        '@angular-eslint/template/no-distracting-elements': 'error',
+
+        '@angular-eslint/template/no-duplicate-attributes': 'error'
     }
 };
--- a/angular.js
+++ b/angular.js
@@ -32,5 +32,41 @@ module.exports = {
             Overrides to Angular extra recommended rules:
             https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/src/configs/recommended--extra.json
         */
+
+        /*
+            Overrides to Angular rules outside of the recommended configuration:
+        */
+
+        /*
+            Extract templates and styles into a separate files. In rare cases, allow for inline templates of a few
+            elements with attributes and three minimal CSS declarations.
+        */
+        '@angular-eslint/component-max-inline-declarations': ['error', { animations: 15, styles: 15, template: 15 }],
+
+        '@angular-eslint/contextual-decorator': 'error',
+
+        /*
+            Do not require a directive to select an element or attribute exclusively, because there are valid use cases
+            for both.
+        */
+        '@angular-eslint/directive-selector': 'off',
+
+        '@angular-eslint/no-attribute-decorator': 'error',
+
+        '@angular-eslint/no-lifecycle-call': 'error',
+        '@angular-eslint/no-pipe-impure': 'error',
+        '@angular-eslint/no-queries-metadata-property': 'error',
+        '@angular-eslint/pipe-prefix': 'error',
+        '@angular-eslint/relative-url-prefix': 'error',
+        '@angular-eslint/use-component-selector': 'error',
+        '@angular-eslint/use-component-view-encapsulation': 'error',
+
+        /*
+            Provide root services with the application root injector in the Injectable decorator. However, NgModule
+            providers are frequently preferred for non-root, module scoped services in applications where tree-shaking
+            is usually irrelevant. Additionally, libraries commonly export services from modules in order to manage
+            dependencies. Consider enabling this rule for libraries to ensure proper tree-shaking when appropriate.
+        */
+        '@angular-eslint/use-injectable-provided-in': 'off'
     }
 };


angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
             Overrides to Angular extra recommended rules:
             https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/src/configs/recommended--extra.json
         */
+
+        /*
+            Overrides to Angular rules outside of the recommended configuration:
+        */
+
+        '@angular-eslint/component-max-inline-declarations': 'error',
```
TrevorKarjanis, line 40: [@angular-eslint/component-max-inline-declarations](http://codelyzer.com/rules/component-max-inline-declarations)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/component-selector': 'error',
```
TrevorKarjanis, line 41: [@angular-eslint/component-selector](http://codelyzer.com/rules/component-selector)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/contextual-decorator': 'error',
```
TrevorKarjanis, line 42: [@angular-eslint/contextual-decorator](http://codelyzer.com/rules/contextual-decorator)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/directive-selector': 'error',
```
TrevorKarjanis, line 43: [@angular-eslint/directive-selector](http://codelyzer.com/rules/directive-selector)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/no-attribute-decorator': 'error',
```
TrevorKarjanis, line 44: [@angular-eslint/no-attribute-decorator](http://codelyzer.com/rules/no-attribute-decorator)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/no-forward-ref': 'error',
```
TrevorKarjanis, line 45: [@angular-eslint/no-forward-ref](http://codelyzer.com/rules/no-forward-ref)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/no-input-prefix': 'error',
```
TrevorKarjanis, line 46: [@angular-eslint/no-input-prefix](http://codelyzer.com/rules/no-input-prefix)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/no-lifecycle-call': 'error',
```
TrevorKarjanis, line 47: [@angular-eslint/no-lifecycle-call](http://codelyzer.com/rules/no-lifecycle-call)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/no-pipe-impure': 'error',
```
TrevorKarjanis, line 48: [@angular-eslint/no-pipe-impure](http://codelyzer.com/rules/no-pipe-impure)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/no-queries-metadata-property': 'error',
```
TrevorKarjanis, line 49: [@angular-eslint/no-queries-metadata-property](http://codelyzer.com/rules/no-queries-metadata-property)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/pipe-prefix': 'error',
```
TrevorKarjanis, line 50: [@angular-eslint/pipe-prefix](http://codelyzer.com/rules/pipe-prefix)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/prefer-on-push-component-change-detection': 'error',
```
TrevorKarjanis, line 51: [@angular-eslint/prefer-on-push-component-change-detection](http://codelyzer.com/rules/prefer-on-push-component-change-detection)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/prefer-output-readonly': 'error',
```
TrevorKarjanis, line 52: [@angular-eslint/prefer-output-readonly](http://codelyzer.com/rules/prefer-output-readonly)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/relative-url-prefix': 'error',
```
TrevorKarjanis, line 53: [@angular-eslint/relative-url-prefix](http://codelyzer.com/rules/relative-url-prefix)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/sort-ngmodule-metadata-arrays': 'error',
```
TrevorKarjanis, line 54: [@angular-eslint/sort-ngmodule-metadata-arrays](https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/docs/rules/sort-ngmodule-metadata-arrays.md)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/use-component-selector': 'error',
```
TrevorKarjanis, line 55: [@angular-eslint/use-component-selector](http://codelyzer.com/rules/use-component-selector)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/use-component-view-encapsulation': 'error',
```
TrevorKarjanis, line 56: [@angular-eslint/use-component-view-encapsulation](http://codelyzer.com/rules/use-component-view-encapsulation)

angular.js
```diff
@@ -32,5 +32,28 @@ module.exports = {
...
+        '@angular-eslint/use-injectable-provided-in': 'error'
```
TrevorKarjanis, line 57: [@angular-eslint/use-injectable-provided-in](http://codelyzer.com/rules/use-injectable-provided-in)

TrevorKarjanis (commented):
Please leave a comment in the thread for rules you wish to discuss. Click "Load more..." to load hidden comments.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
             Overrides to Angular template recommended rules:
             https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/src/configs/recommended.json
         */
+
+        /*
+            Overrides to Angular rules outside of the recommended configuration:
+        */
+
+        '@angular-eslint/template/accessibility-alt-text': 'error',
+        '@angular-eslint/template/accessibility-elements-content': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: This requires that the [a, button, h1, h2, h3, h4, h5, and h6](https://github.com/angular-eslint/angular-eslint/blob/c8d2321181576152d307fc9fd4c836fdc28915fb/packages/eslint-plugin-template/src/rules/accessibility-elements-content.ts#L40) elements have content or an appropriate alternative attribute.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/accessibility-label-for': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: Requires that labels are associated with a form control with the `for` attribute. We don't require this, and we don't like using IDs. We also have cases where the label is used outside of a form, like a menu.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/accessibility-label-has-associated-control': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: Requires that labels are associated with a form control with the for attribute. We don't require this, and we don't like using IDs. We also have cases where the label is used outside of a form, like a menu. [Deprecates](https://github.com/angular-eslint/angular-eslint/pull/392) `accessibility-label-for`.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
             Overrides to Angular template recommended rules:
             https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/src/configs/recommended.json
         */
+
+        /*
+            Overrides to Angular rules outside of the recommended configuration:
+        */
+
+        '@angular-eslint/template/accessibility-alt-text': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: This requires that the [img, area, object, and input elements](https://github.com/angular-eslint/angular-eslint/blob/c8d2321181576152d307fc9fd4c836fdc28915fb/packages/eslint-plugin-template/src/rules/accessibility-alt-text.ts#L33) have alt text.

I'm not aware of any projects that currently have ARIA accessibility requirements, and projects that don't wouldn't to enforce them. If that is the common case, then I would propose leaving these rules disabled even if they don't result in errors in preliminary testing. I've left a comment with notes following testing each rule, and given those results, I would propose the following unless this is not actually common.

**Proposal**
Leave the following rules disabled.
- `@angular-eslint/template/accessibility-alt-text`
- `@angular-eslint/template/accessibility-elements-content`
- `@angular-eslint/template/accessibility-label-for`
- `@angular-eslint/template/accessibility-label-has-associated-control`

**Question**
[template-accessibility-table-scope](http://codelyzer.com/rules/template-accessibility-table-scope/) and [template-accessibility-valid-aria](http://codelyzer.com/rules/template-accessibility-valid-aria/) validate correct usage of ARIA attributes. Do we leave these disabled, because we don't expect their use?

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/accessibility-elements-content': 'error',
+        '@angular-eslint/template/accessibility-label-for': 'error',
+        '@angular-eslint/template/accessibility-label-has-associated-control': 'error',
+        '@angular-eslint/template/accessibility-table-scope': 'error',
+        '@angular-eslint/template/accessibility-valid-aria': 'error',
+        '@angular-eslint/template/click-events-have-key-events': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: We don't currently require providing a key alternative to click events.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/conditional-complexity': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: This is worth considering. I follow this rule as a personal practice, because I don't like code in templates. I have definitely seen more complex conditionals in reviews, but I'm not sure anything more than the default for this rule, 4.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/cyclomatic-complexity': 'error',
+        '@angular-eslint/template/i18n': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: This with the `check-text` option is a great rule for projects that are localize or are internationalizing for the possibility. I'm not sure it should be enabled by default, though.

angular-template.js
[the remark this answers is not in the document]
jattasNI: We discussed our philosophy around accessibility in our meeting today. I have an action item to reach out to some software leaders to discuss where it fits in our priority list and whether shared libraries should allow for accessible apps or or do what they can to require accessible apps.

angular.js
TrevorKarjanis: Similarly to sort-vars, sort-keys, sort-imports and sort-type-union-intersection-members, sorting array members alphabetically is not a generally expected practice. Although, some like myself may do so anyway.

**Proposal**
Leave the rule disabled.

angular.js
TrevorKarjanis: This rule is interesting. You have explicitly disable view encapsulation, so it's almost certainly intentional. I don't know if this is for environments where they're having trouble with people disabling encapsulation?

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/mouse-events-have-key-events': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: This is an accessibility requirement.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/no-any': 'error',
+        '@angular-eslint/template/no-autofocus': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: This is an accessibility requirement.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/no-call-expression': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: This is over kill in our case. Yes, we need to consider performance and use memoization and pipes in many cases. However, there are also many other cases where a function call has no impact on performance—where a 1ms execution is imperceivable for users.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/no-distracting-elements': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: This is a **visual** accessibility requirement and should require additional evaluation.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/no-duplicate-attributes': 'error',
+        '@angular-eslint/template/no-positive-tabindex': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: This is a **Google** requirement for accessibility.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/use-track-by-function': 'error'
```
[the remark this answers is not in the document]
TrevorKarjanis: This likely applies to certain applications.

**Proposal**
Leave this rule disabled.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
             Overrides to Angular template recommended rules:
             https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/src/configs/recommended.json
         */
+
```
jattasNI, line 10: We discussed the accessibility rules in our meeting today. Our decision was to disable all of them for now in our recommended config. This (sadly) most accurately reflects where the company is today with accessibility--it's not something that developers should spend time enforcing.

There is an upcoming push from the Design team and the design system to start to care more. At that time we can revisit this decision and hopefully turn the rules on with the backing of leadership to spend developer time making apps accessible.

@TrevorKarjanis can you update the PR to reflect this with a comment something like this?
"At the moment, there's no NI-wide requirement to enforce accessible apps. These rules can be useful if individual apps choose to prioritize it. They may be enabled in the future as the priority of accessibility increases."

TrevorKarjanis: Yes, and I would also recognize that this is already reflected in our JavaScript and TypeScript configurations.

angular.js
TrevorKarjanis: Worthy of consideration, but I would increase the defaults.

**Proposal**
```javascript
'component-max-inline-declarations': [true, {"animations": 15, "styles": 15, "template": 15}]
```

angular.js
TrevorKarjanis: This rule requires configuration that is specific to an application.

**Proposal**
Leave this rule disabled, but add it to section mentioning rules to consider.

angular.js
TrevorKarjanis: This rule forces a directive to be an element or attribute which I don't think is appropriate in general. It also requires application specific configuration. 

**Proposal**
Leave this rule disabled.

angular.js
TrevorKarjanis: ForwardRef is only used for specific, necessary problems.

**Proposal**
Leave this rule disabled.

angular.js
TrevorKarjanis: I avoid verbs when naming @Inputs, but we don't have a general convention to apply this rule with.

**Proposal**
Leave this rule disabled.

angular.js
TrevorKarjanis: This rule requires application specific configuration.

**Proposal**
Leave this rule disabled, but include it in a section for rules to consider in the README.

angular.js
TrevorKarjanis: This rule is for specific applications.

**Proposal**
Leave this rule disabled.

angular.js
TrevorKarjanis: This is an opinion on the use of readonly which I have not seen nor want to embark on a general rule. This is probably better handled in the TypeScript configuration.

**Proposal**
Leave this rule disabled.

angular.js
jattasNI: One nice thing about causing it to be an error is that it forces devs to add a comment suppressing the rule, which helps highlight that a component has diverged from the normal encapsulation behavior. And if your codebase has a policy of requiring an explanation when suppressing rules, then that explanation can serve as documentation for why it's diverging.

Anyway, I agree to keep it an error, just speculating.

angular-template.js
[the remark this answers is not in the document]
jattasNI: I agree it seems like a useful rule. Improve readability and testability by moving complex code out of the template and into the component TS.

angular-template.js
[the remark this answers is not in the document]
jattasNI: Can you provide an example of a violation of this rule? The docs don't have any and I don't think I understand what they mean by "Disallows calling expressions in templates, except for output handlers". Aren't all function calls, property accesses, operators, etc known as "expressions"? Does this mean your template can only be dumb static code? Clearly I'm missing something.

Is any of this code illegal?
```html
<div *ngIf="step === 'amount' || (step === 'card' && transaction.action !== 'iframe-sent' && transaction.action !== 'html-provided' && transaction.action !== 'redirect-to-url')">
  Enter your card details
</div>
<div *ngIf="canShowCardStep">
  Enter your card details
</div>
```

angular-template.js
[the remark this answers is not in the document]
jattasNI: Wow, the docs are terrible. "Ensures trackBy function is used. Rationale: The use of ‘trackBy’ is considered a good practice." 🙄

This [blog](https://netbasal.com/angular-2-improve-performance-with-trackby-cc147b5104e5) gives a decent summary. If you're using an `ngFor` to create elements, when the collection you're iterating changes Angular can't tell which items were added/removed so it has to recreate all the elements. A `trackBy` function lets you provide a unique identifier for each item so Angular can reuse elements.

I think this could be a useful optimization but agree we shouldn't require it for every `ngFor`.

angular.js
jattasNI: Any particular rationale for increasing the defaults? I like the consistency of being able to find styles and templates in their own files and would only use inline declarations for extremely trivial cases (1 or 2 lines, which aligns well with the default of `3` for `styles` and `template`).

**Proposal**: `        '@angular-eslint/component-max-inline-declarations': 'error',`

angular.js
jattasNI: I agree that it doesn't make sense to enforce a specific prefix for component selectors from our shared ruleset.

It would be nice if it were possible to turn off that check but still enforce kebab-case and element selectors over attribute selectors. The codelyzer docs don't show how you'd do that but the [docs on angular-eslint GitHub](https://github.com/angular-eslint/angular-eslint/blob/ecbe68431fb73177d905676fef3df9be9c646636/packages/eslint-plugin/docs/rules/component-selector.md) say it *might* be possible. But even if we did that, apps would still have to override the entire config to provide a selector so including the rule in the shared ruleset doesn't add much benefit.

I think I've come around to agreeing with your proposal but wanted to lay out the landscape for others to comment.

angular-template.js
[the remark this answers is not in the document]
TrevorKarjanis: **Proposal**
```javascript
'@angular-eslint/template/conditional-complexity': 'error'
```

angular-template.js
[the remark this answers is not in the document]
TrevorKarjanis: They're referring function calls, likely because these hide complexity and tend to be more complex. However, if you're like me you'd put your first NgIf into a function call, and it would still execute at <1ms. I keep track of the number of conditionals and try to make these reactive with setters, and I definitely draw the line at array iteration.

The second primary factor to CD performance is the number of components on the page. This is why we added support  to material side-navs for lazy-loading and full page viewing ([not full and animations broken](https://stackblitz.com/edit/angular-qaqyoc)). This is a trick I stole from mobile, like iOS, which allows you to build vertical navigation without routing, destroy invisible views, and drop your component count to nothing FOR FREE. This is why I hate side-nav implementations that use floats to try to be performant. It's not, and they're missing the point (bit of rambling—super interesting topic).

angular-template.js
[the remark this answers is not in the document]
TrevorKarjanis: Every NgFor tracks by reference by default. Therefore, it can tell what was added and removed if the references weren't changed. This isn't so helpful if your array is populated by a request that creates new references.

angular.js
TrevorKarjanis: Yes, I do inline by default in StackBlitz, and so I've gotten a lot of time with it. You can't do anything with three lines of CSS. Three lines of HTML is annoying when you need four, five, or six is typical for three lines per element. The class probably won't be much either, so the rationale is that ~45 lines fits nicely when you should have already considered separate files anyway. If you want to bring it down significantly, I would say 15, 15, and 6 for two elements or attributes with line length.

angular-template.js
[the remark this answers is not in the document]
jattasNI: Comment copied from below because I forgot this thread had already started:
We discussed the accessibility rules in our meeting today. Our decision was to disable all of them for now in our recommended config. This (sadly) most accurately reflects where the company is today with accessibility--it's not something that developers should spend time enforcing.

There is an upcoming push from the Design team and the design system to start to care more. At that time we can revisit this decision and hopefully turn the rules on with the backing of leadership to spend developer time making apps accessible.

Trever will update the PR to reflect this with a comment something like this:
"At the moment, there's no NI-wide requirement to enforce accessible apps. These rules can be useful if individual apps choose to prioritize it. They may be enabled in the future as the priority of accessibility increases."

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        /*
+            Overrides to Angular rules outside of the recommended configuration:
+        */
+
+        '@angular-eslint/template/accessibility-alt-text': 'error',
+        '@angular-eslint/template/accessibility-elements-content': 'error',
+        '@angular-eslint/template/accessibility-label-for': 'error',
+        '@angular-eslint/template/accessibility-label-has-associated-control': 'error',
+        '@angular-eslint/template/accessibility-table-scope': 'error',
```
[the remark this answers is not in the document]
TrevorKarjanis: Resolved as part of [comment r652709294](https://github.com/ni/javascript-styleguide/pull/45/files#r652709294).

angular-template.js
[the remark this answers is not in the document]
TrevorKarjanis: Resolved as part of [comment r652709294](https://github.com/ni/javascript-styleguide/pull/45/files#r652709294).

angular-template.js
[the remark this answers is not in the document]
TrevorKarjanis: Resolved as part of [comment r652709294](https://github.com/ni/javascript-styleguide/pull/45/files#r652709294).

angular-template.js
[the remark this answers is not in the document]
TrevorKarjanis: '@angular-eslint/template/mouse-events-have-key-events': 'error',

angular-template.js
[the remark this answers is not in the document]
TrevorKarjanis: Resolved as part of [comment r652709294](https://github.com/ni/javascript-styleguide/pull/45/files#r652709294).

angular.js
TrevorKarjanis: This usually will be disabled for libraries. Tree-shaking is generally irrelevant for applications, and we actually prefer to use NgModule providers for services that are not in root.

angular-template.js
```diff
@@ -7,5 +7,28 @@ module.exports = {
...
+        '@angular-eslint/template/accessibility-valid-aria': 'error',
```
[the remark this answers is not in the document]
jattasNI: Note from Trevor: this just validates the use of some attributes, doesn't force an app to use them. So we're keeping it on despite turning off the other accessibility rules.

angular-template.js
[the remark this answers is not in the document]
jattasNI: We agreed in person that this rule makes sense to turn on by default. It helps make sure that apps think about localization from the beginning. An app can add `i18n` tags without setting up all localization tooling so it's not too big a burden to have it on when setting up new projects.

Let's enable this with the `check-text` option and a comment something like this: "This rule helps enforce that user-visible strings are marked for localization. We recommend enabling it for new apps if there's any chance that they'll be localized in the future and for existing apps that are localized. It's acceptable to disable for existing apps that aren't currently localized and new apps that will never be localized."

angular-template.js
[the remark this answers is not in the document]
jattasNI: @TrevorKarjanis this is ready to close

angular-template.js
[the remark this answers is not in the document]
jattasNI: @TrevorKarjanis this is ready to close

angular-template.js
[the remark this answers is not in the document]
jattasNI: @TrevorKarjanis this is ready to close

angular.js
jattasNI: @TrevorKarjanis this is ready to close

angular.js
jattasNI: We discussed this in person. @TrevorKarjanis could you see what happens if we enforce `kebab-case` but leave `prefix` as empty? That seems preferable if it works.

angular.js
jattasNI: We discussed this rule in person and didn't really understand what it enforced regarding element vs attribute. Is it saying that directives can only apply to elements or attributes but not both? If so we agree that seems inappropriate.

Similar to `component-selector`, if it's possible to enforce just `kebab-case` then that would be nice-to-have.

@TrevorKarjanis  can you help us understand what's possible?

angular.js
jattasNI: @mure this thread is awaiting your vote

angular.js
jattasNI: @mure this thread is awaiting your vote

jattasNI: @TrevorKarjanis this thread is ready to close

angular.js
jattasNI: @mure this thread on @angular-eslint/sort-ngmodule-metadata-arrays is awaiting your vote

angular.js
jattasNI: @rajsite This thread on @angular-eslint/use-component-view-encapsulation is awaiting your vote.

angular.js
TrevorKarjanis: Yes, I will give it a try.

TrevorKarjanis: The rule simply doesn't work without providing a prefix.

angular.js
TrevorKarjanis: For reasons described in the rule documentations as well as the [style guide](https://angular.io/guide/styleguide#style-02-06), attribute directives are camel case. Yes, though, this enforces that Directives apply to either elements or attributes of elements, e.g. `app-root` or `app-root[appDirective]`. I'm not sure either is more common, because both are excellent for extending the functionality. Element directives simply apply to all elements of a certain type. Again, this rule doesn't work without a prefix.

angular.js
TrevorKarjanis: I'm willing and able to act as @mure 's proxy vote.

angular.js
jattasNI: We agreed in person to disable this since it's not universally applicable. @TrevorKarjanis we can close this thread.

angular.js
jattasNI: @TrevorKarjanis this is ready to close.

angular.js
jattasNI: @TrevorKarjanis this is ready to close.

angular.js
jattasNI: An alternate take is that we should try to be as strict as possible to express our intention more clearly, even if we doubt someone would ever bother to try to assign to an output.

**Proposal**: leave this rule enabled (assuming it has an autofixer; if it doesn't, I agree it would be annoying to fix manually)

jattasNI: @LukasKall this thread has a new proposal for you to consider. Let us know if you change your vote.

jattasNI: If we end up 50-50, @TrevorKarjanis gets to be the deciding vote.

angular.js
jattasNI: @TrevorKarjanis this thread is ready to close

angular.js
jattasNI: @TrevorKarjanis this is ready to close

angular.js
mure: I consent

angular.js
jattasNI: @TrevorKarjanis this is ready to close

README.md
```diff
@@ -88,6 +88,9 @@ configure ESLint for Angular projects** especially when migrating from TSLint. [
     ```bash
     ng g @angular-eslint/schematics:convert-tslint-to-eslint <PROJECT NAME>
     ```
+4. Enable these application specific rules as appropriate.
```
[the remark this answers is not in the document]
TrevorKarjanis: Good point, there are many cases where this isn't necessary.

PR #46, opened 2021-06-16
Evaluate diverged Angular rules.

TrevorKarjanis:
- Add a log to evaluate diverged Angular rules to print-evaluated-rules script.

Linted and tested
- Verified that diverged rules are printed
- Verified that overridden rules are not printed

### DIFF
--- a/tools/print-evaluated-rules.js
+++ b/tools/print-evaluated-rules.js
@@ -1,4 +1,7 @@
 const { ESLint } = require('eslint');
+const angularExtraRules = require('@angular-eslint/eslint-plugin/dist/configs/recommended--extra.json').rules;
+const angularRules = require('../angular').rules;
+const angularTemplateRules = require('../angular-template').rules;
 
 (async () => {
     const hasWarn = function (config) {
@@ -11,6 +14,19 @@ const { ESLint } = require('eslint');
     const configTypescriptTypechecked = await eslint.calculateConfigForFile(`${__dirname}/../test/typescript-type-checking/index.js`);
     const configAngular = await eslint.calculateConfigForFile(`${__dirname}/../test/angular/index.ts`);
 
+    const getDivergedRules = rules => (
+        Object.keys(rules)
+            .filter(key => !key.startsWith('@angular-eslint'))
+            .filter(key => !configTypescript.rules[key] && !configTypescriptTypechecked.rules[key])
+            .reduce((config, key) => {
+                config[key] = rules[key];
+                return config;
+            }, {})
+    );
+    const angularDivergedRules = Object.assign(getDivergedRules(angularRules), getDivergedRules(angularExtraRules));
+    const angularTemplateDivergedRules = getDivergedRules(angularTemplateRules);
+    const angularHasDivergedRules = !!Object.keys(angularDivergedRules).length || !!Object.keys(angularTemplateDivergedRules).length;
+
     global.console.log('-------- Evaluated rules:');
     global.console.log('Evaluated ESLint rules:');
     global.console.log(JSON.stringify(configEslint, undefined, 4));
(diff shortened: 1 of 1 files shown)


tools/print-evaluated-rules.js
```diff
@@ -10,6 +11,12 @@ const { ESLint } = require('eslint');
     const configTypescript = await eslint.calculateConfigForFile(`${__dirname}/../test/typescript/index.js`);
     const configTypescriptTypechecked = await eslint.calculateConfigForFile(`${__dirname}/../test/typescript-type-checking/index.js`);
     const configAngular = await eslint.calculateConfigForFile(`${__dirname}/../test/angular/index.ts`);
+    const configAngularExtra = Object.keys(angularExtra.rules)
```
rajsite, line 14: This variable doesn't represent the evaluated config like the others. Maybe a name like `angularExtraRulesNotOverridden`?

tools/print-evaluated-rules.js
```diff
@@ -20,6 +27,8 @@ const { ESLint } = require('eslint');
     global.console.log(JSON.stringify(configTypescriptTypechecked, undefined, 4));
     global.console.log('Evaluated Angular rules:');
     global.console.log(JSON.stringify(configAngular, undefined, 4));
+    global.console.log('Evaluated diverged Angular rules:');
```
rajsite, line 30: The check isn't for all Angular rules, just the subset that is in Angular extra. Also diverged sounds like it might compare the value of the overwritten rules but it just checks if it is overwritten.
Maybe `'Angular extra rules not overwritten:'`?

tools/print-evaluated-rules.js
```diff
@@ -20,6 +27,8 @@ const { ESLint } = require('eslint');
...
+    global.console.log(JSON.stringify(configAngularExtra, undefined, 4));
 
     global.console.log('-------- Audits:');
     global.console.log(`Evaluated ESLint rules has warn?: ${hasWarn(configEslint)}`);
```
rajsite, line 34: Can we add an audit line for these changes? maybe `Has Angular extra rules that are not overwritten?: `

rajsite (changes requested):

tools/print-evaluated-rules.js
TrevorKarjanis: I would definitely prefer to continue with the term diverged, because it best describes the purpose of the check. Given that, I will include any overrides defined in `angular.ts` and `angular-template.ts`. I didn't consider that initially, because it's a bit of an awkward setup through no fault of our own. I'm not concerned about the values, because if they're equal then they're not divergent and we wouldn't define an override.

tools/print-evaluated-rules.js
TrevorKarjanis: Yes.

tools/print-evaluated-rules.js
```diff
@@ -11,6 +14,19 @@ const { ESLint } = require('eslint');
...
+    const getDivergedRules = rules => (
+        Object.keys(rules)
+            .filter(key => !key.startsWith('@angular-eslint'))
+            .filter(key => !configTypescript.rules[key] && !configTypescriptTypechecked.rules[key])
+            .reduce((config, key) => {
+                config[key] = rules[key];
+                return config;
+            }, {})
+    );
+    const angularDivergedRules = Object.assign(getDivergedRules(angularRules, getDivergedRules(angularExtraRules)));
```
rajsite, line 26: Is this accurate? `getDivergedRules(angularRules, getDivergedRules(angularExtraRules))` is passing a second argument to getDivergedRules which does not seemed to be used by the function

rajsite (changes requested):

tools/print-evaluated-rules.js
TrevorKarjanis: Fixed!

rajsite (approved):

Issue #43, opened 2021-05-27
@typescript-eslint/no-floating-promises conflicts with no-void


LukasKall:
The `@ni/eslint-config/typescript-requiring-type-checking` rule set enables [@typescript-eslint/no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md) which is in general desired to avoid unhandled promises. In some cases, you want to explicitly mark a promise as intentionally not awaited. You can add `void` to indicate that.

```javascript
void Promise.reject('value');
```

But this conflicts with [no-void](https://eslint.org/docs/rules/no-void), which is enabled as well. It has an option [allowAsStatement](https://eslint.org/docs/rules/no-void#allowasstatement) which allows this use of void and it's suggested by the [@typescript-eslint/no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/v4.25.0/packages/eslint-plugin/docs/rules/no-floating-promises.md#ignorevoid) rule to enable it.
> With this option set to true, and if you are using no-void, you should turn on the allowAsStatement option.

I propose to enable the allowAsStatement option on the no-void rule. We can either do this for all the rule sets (JavaScript and TypeScript), or just for the TypeScript ones. I'm curious what other opinions are.

rajsite:
@LukasKall do you have some example code snippets that show some context about why the promise result should be ignored? We already have one exception listed for Angular tests that rely on [jasminewd2](https://github.com/ni/javascript-styleguide/blob/55785af7da36361729f53beef7093d09cb53add8/typescript-requiring-type-checking.js#L14) listed which results in disabling the rule for tests.

LukasKall:
1. I want to show an error message as a dialog using an async function. But I don't want to wait for the user to confirm the error message.
```typescript
void showErrorMessage(
    'Invalid configuration',
    'Please provide a valid Analysis Automation Procedure configuration.'
);
```
2. When you want to fetch data in an `useEffect` React hook, you have to use an async function. But the `useEffect` hook doesn't accept an async function. So you have to call it inside the hook without awaiting it.
```typescript
useEffect(() => {
    const fetchData = async (): Promise<void> => {
        // doing async calls in here with await
    };
    void fetchData();
}, []);
```
3. Sometimes you want to call an asnyc function in a React/HTML event handler, e.g. `onClick`. But you cannot await it because the event handler doesn't accept an async function.
```typescript
<button
[...]

PR #51, opened 2021-06-30
Allow void as statement

LukasKall:
Allow `void` as statement to explicitly mark a promise as intentionally not awaited, as discussed in issue #43.

@rajsite I think we should publish a new minor release after this. It's not a breaking change, but it changes a rule to be less strict.


mure (approved):

rajsite:
@LukasKall it looks like this change includes package-lock changes which are unexpected. My guess is that you have npm@7 installed locally and running npm install resulted in updates to the lockfile to the npm7 format (aka v2). We should not include that change in this PR and explicitly require update to npm7.

Can you either remove the lockfile change in the PR or temporarily switch back to `npm i -g npm@6` for this PR?

rajsite (changes requested):
lock file changes should not be included

LukasKall:
> @LukasKall it looks like this change includes package-lock changes which are unexpected. My guess is that you have npm@7 installed locally and running npm install resulted in updates to the lockfile to the npm7 format (aka v2). We should not include that change in this PR and explicitly require update to npm7 in a separate PR.
> 
> Can you either remove the lockfile change in the PR or temporarily switch back to `npm i -g npm@6` for this PR?

@rajsite  Sorry, my bad. I reverted the changes to the lockfile.

rajsite (approved):

PR #53, opened 2021-07-14
Evaluate base TSConfig compiler options

mure:
We can use this PR to evaluate and vote on a recommended set of compiler flags for a base TSConfig ([info and full list of options](https://www.typescriptlang.org/tsconfig)). 
A couple things to note:
- Most options in a TSConfig are environment or project specific, such as the list of files to include or the module system. This base config only includes recommended flags (according to the TS docs) and type checking rules that weren't already covered by linter rules. If we decide in the future that we want to provide, say, a recommended config for node projects, we can add another file in the `configs` directory that extends from this base.
- Some of the flags are from pretty recent versions of TypeScript. Should we add a peer dependency to this package for typescript?


LukasKall (approved):
I applied the options in one of my projects which is already using our TypeScript rules. I could still compile the code without any modifications. So I guess the compiler options align pretty well with our style guide and I'm fine with using these.

configs/base.json
```diff
@@ -0,0 +1,12 @@
+{
+    "compilerOptions": {
+        "esModuleInterop": true,
+        "forceConsistentCasingInFileNames": true,
+        "noImplicitOverride": true,
+        "noImplicitReturns": true,
+        "noPropertyAccessFromIndexSignature": true,
+        "noUncheckedIndexedAccess": true,
+        "skipLibCheck": true,
```
jattasNI, line 9: [`skipLibCheck`](https://www.typescriptlang.org/tsconfig#skipLibCheck) is the only one that isn't a slam dunk for me. It turns of type checking of .d.ts files in node_modules as a performance optimization, which can cause problems if you pull in two different versions of the same library.

Do we have a sense for how common this problem is or how much this improves performance? In the absence of any data, I'd vote to keep stricter checks (i.e. `"skipLibCheck": false`) but could be swayed if we have experience with this.

configs/base.json
```diff
@@ -0,0 +1,12 @@
+{
```
jattasNI, line 1: As part of this PR we should include updates to the README.md explaining how to include this tsconfig in a project.

configs/base.json
TrevorKarjanis: I wouldn't think pulling two different versions of a library is common for front-end projects—maybe Node.js? Doesn't npm install conflicts as a subdirectory of the dependent?

jattasNI: We discussed in person. There was [some concern](https://stackoverflow.com/questions/66817188/how-to-use-declare-an-external-module-which-wont-pass-strict-and-noimplicitr) that setting this to `false` would cause our strict tsconfig to start reporting errors on type declarations of libraries that use looser compiler configurations. This would be problematic because there wouldn't be ways to work around it, especially if your app wasn't even using the signature that was weakly typed. 

@mure  could you experiment and see if our understanding of the behavior is correct?

mure: I did some experiments and wasn't able to find any ways that having `skipLibCheck: false` would cause problems importing libraries with looser config options. We decided in person that we should go ahead and keep it false. We have anecdotal evidence that having it off doesn't cause any noticeable performance issues (from the nimble repo).

rajsite:
Gonna close due to lack of activity but we should continue when we have the chance!

PR #58, opened 2021-09-02
Dependency update and a few new rules

rajsite:
Overview:
- Updated dependencies to latest versions. Used the [development scripts](https://github.com/ni/javascript-styleguide/blob/main/CONTRIBUTING.md#development-scripts) to evaluate rule changes manually.
- Formatted js files to align in style (lines between rules, alphabetical order)
- Modified the utilities to sort properties and filled some gaps, ie angular template rules warn checks
- Added new rules that have been included in updates with proposed values
- Switched to npm 7 for dev, shouldn't impact package users
- Added documentation concept of rule groups for [accessibility] [application-prefix] [strict-null-checks]
- Added documentation for all disabled rules that have been evaluated

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,13 +135,13 @@ Text search for the tag associated with a specific rule group in the repository 
 
 #### Application prefix
 
-Tag: [`[application-prefix]`](https://github.com/ni/javascript-styleguide/search?q=%5Bapplication-prefix%5D)
+Tag: [`[application-prefix]`](https://github.com/ni/javascript-styleguide/search?q=%5Bapplication-prefix%5D&type=code)
 
 Prefixes are generally added to named objects such as the selector for Components in Angular applications. Projects should consider enabling this rule group so that names can be consistently prefixed making them easier to share between applications and to minimize the chance of conflicts when using shared libraries.
 
 #### Strict null checks
 
-Tag: [`[strict-null-checks]`](https://github.com/ni/javascript-styleguide/search?q=%5Bstrict-null-checks%5D)
+Tag: [`[strict-null-checks]`](https://github.com/ni/javascript-styleguide/search?q=%5Bstrict-null-checks%5D&type=code)
 
 When `strictNullChecks` are enabled the values `null` and `undefined` are treated as distinct types by the compiler. For example, with `strictNullChecks` enabled, the value `null` could not be directly assigned to a binding of a `Cat` object, ie `const cat: Cat = null` would be a compile error. The `null` value is a distinct type and the binding would have to explicitly state that it can have a `null` value, ie `const cat: Cat | null = null;`.
 
@@ -153,7 +153,7 @@ However, we encourage new applications to leverage `strictNullChecks` for develo
 
 #### Accessibility
 
-Tag: [`[accessibility]`](https://github.com/ni/javascript-styleguide/search?q=%5Baccessibility%5D)
+Tag: [`[accessibility]`](https://github.com/ni/javascript-styleguide/search?q=%5Baccessibility%5D&type=code)
 
 There currently isn't an NI organization wide requirement to enforce accessibility in applications. The rule group should be enabled if individual applications prioritize accessibility.
 
--- a/README.md
+++ b/README.md
@@ -88,9 +88,7 @@ configure ESLint for Angular projects** especially when migrating from TSLint. [
     ```bash
     ng g @angular-eslint/schematics:convert-tslint-to-eslint <PROJECT NAME>
     ```
-4. Enable these application-specific rules if your application uses a specific prefix for components or pipes.
-    - [@angular-eslint/component-selector](http://codelyzer.com/rules/component-selector)
-    - [@angular-eslint/pipe-prefix](http://codelyzer.com/rules/pipe-prefix)
+4. Evaluate the [project specific rule groups](#evaluate-project-specific-rule-groups) to manually add to your lint configuration. For Angular applications in particular, consider enabling the [`[application-prefix]`](#application-prefix) rule group.
 
 ## Usage
 
@@ -135,7 +133,13 @@ Several sets of rules may be enabled based on requirements of a given project. B
 
 Text search for the tag associated with a specific rule group in the repository to find the related rules. If enabling a rule group, the rules should be toggled from `'off'` to `'error'` unless the rule comment says otherwise.
 
-#### Strict null checking
+#### Application prefix
+
+Tag: `[application-prefix]`
+
+Prefixes are generally added to named objects such as the selector for Components in Angular applications. Projects should consider enabling this rule group so that names can be consistently prefixed making them easier to share between applications and to minimize the chance of conflicts when using shared libraries.
+
+#### Strict null checks
 
 Tag: `[strict-null-checks]`
 
@@ -195,7 +199,7 @@ ESLint offers several [ways to disable a rule for a line or file](https://eslint
 
 Modern IDEs can be configured to provide live feedback about ESLint errors.
 
-**Visual Studio Code**
+#### Visual Studio Code
 
 Install the [ESLint Extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint).
 
@@ -213,7 +217,7 @@ You can [configure a repository to prompt developers to install this extension](
 
 Follow the `@angular-eslint` [instructions](https://github.com/angular-eslint/angular-eslint#linting-html-files-and-inline-templates-with-the-vscode-extension-for-eslint) for linting HTML files and inline-templates with Angular.
 
-**JetBrains WebStorm**
+#### JetBrains WebStorm
 
 Follow the [instructions in the WebStorm documentation](https://www.jetbrains.com/help/webstorm/eslint.html#ws_js_eslint_activate) to activate and configure ESLint automatically in the Settings ≫ Preferences dialog.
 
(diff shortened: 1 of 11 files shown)


typescript-requiring-type-checking.js
```diff
@@ -77,6 +77,8 @@ module.exports = {
 
         '@typescript-eslint/no-unnecessary-type-arguments': 'error',
 
+        '@typescript-eslint/no-unsafe-argument': 'error',
```
rajsite, line 80: new rule from update [@typescript-eslint/no-unsafe-argument](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-argument.md)

typescript-requiring-type-checking.js
```diff
@@ -104,6 +106,8 @@ module.exports = {
 
         '@typescript-eslint/prefer-reduce-type-parameter': 'error',
 
+        '@typescript-eslint/prefer-return-this-type': 'error',
```
rajsite, line 109: new rule from update [@typescript-eslint/prefer-return-this-type](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-return-this-type.md)

typescript-requiring-type-checking.js
TrevorKarjanis: Huh, [we didn't know this wasn't in effect](https://github.com/ni-kismet/helium-security-ui/pull/421). It's very important when working with untyped imports.

angular-template.js
```diff
@@ -52,6 +70,13 @@ module.exports = {
 
         '@angular-eslint/template/no-distracting-elements': 'error',
 
-        '@angular-eslint/template/no-duplicate-attributes': 'error'
+        '@angular-eslint/template/no-duplicate-attributes': 'error',
+
+        /*
+            This rule primarily impacts accessibility so currently requires enforcement per project.
+        */
+        '@angular-eslint/template/no-positive-tabindex': 'off',
```
TrevorKarjanis, line 78: Is this really the best strategy? I feel like a team that needs to enable the accessibility rules would prefer them grouped and outlined under the comment.

angular.js
```diff
@@ -53,12 +58,28 @@ module.exports = {
 
         '@angular-eslint/no-attribute-decorator': 'error',
 
+        '@angular-eslint/no-forward-ref': 'off',
```
TrevorKarjanis, line 61: Why is it necessary to declare all the rules? Shouldn't semver address this? It's useful to keep the scope of the ruleset to the stuff we care about.

rajsite: We cared enough to not enable it. Capturing it here shows that we evaluated it and made a decision. Also makes maintenance easier. During updates one can print the properties and see what is not captured here. That's how I found the other two rules that were not enabled during update.

angular-template.js
rajsite: It makes it easier to maintain to have the rules listed and ordered like they print in the tools. We could tag them in a comment like:

```
        /*
            [accessibility] This rule primarily impacts accessibility so currently requires enforcement per project.
        */
        '@angular-eslint/template/no-positive-tabindex': 'off',
```
To make that group of rules easier to identify. I like that idea for also marking rules pertaining to strict null checks, maybe a comment tag like `[strict-null-checks]`.

angular.js
rajsite: It might actually be a good idea to include a comment describing why the rule was not enabled.

LukasKall (approved):

angular-template.js
```diff
@@ -52,6 +64,11 @@ module.exports = {
 
         '@angular-eslint/template/no-distracting-elements': 'error',
 
-        '@angular-eslint/template/no-duplicate-attributes': 'error'
+        '@angular-eslint/template/no-duplicate-attributes': 'error',
+
+        /* [accessibility] */
+        '@angular-eslint/template/no-positive-tabindex': 'off',
+
+        '@angular-eslint/template/use-track-by-function': 'off'
```
jattasNI, line 72:
```suggestion
        /*
            Providing a `trackBy` function in `ngFor` loops can improve performance in specific cases where Angular can't track references, but it's overkill to require it for every `ngFor` so this rule is disabled.
        */
        '@angular-eslint/template/use-track-by-function': 'off'
```

angular.js
jattasNI: I like the rationale of adding comments to document why rules weren't enabled. I added a few notable ones from #45 as suggestions, but I don't think we need to be comprehensive about it in this PR.

jattasNI (approved):

angular-template.js
rajsite: Added rule groups for `[accessibility]`, `[strict-null-checks]`, `[application-prefix]`, and corresponding content to the README.

angular.js
rajsite: Was enough of a push for me to audit the rest. All the existing rules configured to 'off' should have justification and all the rules outside the recommended sets (for typescript and angular) should be captured in their respective config.

We rely on airbnb for the javascript / base rule configuration as evaluating all the base JS rules would be another large task, so not all rules from the base eslint rule set are captured in our ruleset.

README.md
```diff
@@ -129,6 +129,30 @@ Each project's pull request build pipeline should ensure no lint errors can be c
...
+As such, `strictNullChecks` are not recommended by default in order to prevent overhead of rule adoption for existing applications.
+
+However, we encourage new applications to leverage `strictNullChecks` for development. Enabling `strictNullChecks` is the TypeScript compiler recommendation and it is enabled by default in new Angular applications.
+
+#### Accessibility
+
+Tag: `[accessibility]`
+
+There currently isn't an NI organization wide requirement to enforce accessibility in applications. The rule group should be enabled if individual applications prioritize accessibility.
+
```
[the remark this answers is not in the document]
rajsite: Good idea! Added `[application-prefix]`.

rajsite:
@TrevorKarjanis reset you on the PR as there were new rule justifications added and an implementation of the rule groups I proposed in the comments.

README.md
```diff
@@ -129,6 +127,36 @@ Each project's pull request build pipeline should ensure no lint errors can be c
...
+`strictNullChecks` are a powerful tool for code correctness and give us a way to avoid ["The Billion Dollar Mistake"](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/). However, it can be impractical to retrofit `strictNullChecks` configuration into an existing application and requires expanding your mental model for software development for use in new applications.
+
+As such, `strictNullChecks` are not recommended by default in order to prevent overhead of rule adoption for existing applications.
+
+However, we encourage new applications to leverage `strictNullChecks` for development. Enabling `strictNullChecks` is the TypeScript compiler recommendation and it is enabled by default in new Angular applications.
+
+#### Accessibility
+
+Tag: `[accessibility]`
```
jattasNI, line 156:
```suggestion
Tag: [`[accessibility]`](https://github.com/ni/javascript-styleguide/search?q=%5Baccessibility%5D)
```

README.md
```diff
@@ -129,6 +127,36 @@ Each project's pull request build pipeline should ensure no lint errors can be c
...
+
+#### Application prefix
+
+Tag: `[application-prefix]`
+
+Prefixes are generally added to named objects such as the selector for Components in Angular applications. Projects should consider enabling this rule group so that names can be consistently prefixed making them easier to share between applications and to minimize the chance of conflicts when using shared libraries.
+
+#### Strict null checks
+
+Tag: `[strict-null-checks]`
```
jattasNI, line 144:
```suggestion
Tag: [`[strict-null-checks]`](https://github.com/ni/javascript-styleguide/search?q=%5Bstrict-null-checks%5D)
```

README.md
```diff
@@ -129,6 +127,36 @@ Each project's pull request build pipeline should ensure no lint errors can be c
...
 
+### Evaluate project-specific rule groups
+
+Several sets of rules may be enabled based on requirements of a given project. By default the following sets of rules are in an inert / disabled state, but should be evaluated for your integration.
+
+Text search for the tag associated with a specific rule group in the repository to find the related rules. If enabling a rule group, the rules should be toggled from `'off'` to `'error'` unless the rule comment says otherwise.
+
+#### Application prefix
+
+Tag: `[application-prefix]`
```
jattasNI, line 138:
```suggestion
Tag: [`[application-prefix]`](https://github.com/ni/javascript-styleguide/search?q=%5Bapplication-prefix%5D)
```

jattasNI (approved):

README.md
rajsite: I had the exact same thought but with also including `&type=code` in the url parameters. My concern is that GitHub search doesn't include punctuation. See the current [accessibility](https://github.com/ni/javascript-styleguide/search?q=%5Baccessibility%5D) search. The github.dev vscode search does but I couldn't figure out a way to encode a search in URL parameters.

I was on the fence but if you confirm a vote for still including the url (with `&type=code` added) then I'll jump on board too.

jattasNI: Yeah, it's not perfect but I figure there aren't likely to be too many spurious results so it's still valuable.

rajsite:
@mure When you get a chance, the main aspect to review in the PR is the new rules being enabled brought in from version updates.

TrevorKarjanis (approved):

typescript-requiring-type-checking.js
mure: Interesting, I didn't know about the polymorphic 'this' type.

mure (approved):

PR #59, opened 2021-09-08
Monorepo migration

### LINKED ISSUE #41: Angular peer dependencies in non-Angular projects
Since we now have a significant number of dependencies that won't be needed for non-Angular projects, what do you think of updating the "Installation" section of README.md to provide more explicit guidance about using npm view vs npx install-peerdeps?

I'm thinking something like a note in the install-peerdeps section saying "For non-Angular projects, this will install a significant number of unneeded dependencies"

Moved from a [comment](https://github.com/ni/javascript-styleguide/pull/40#discussion_r640652658) from @jattasNI  in [PR#40](https://github.com/ni/javascript-styleguide/pull/40).

rajsite:
# Overview

This PR migrates the repository from a single package repository to a monorepo with multiple packages. One of the primary benefits is projects can install only dependencies they require with npm 7. With npm 7 peerDependencies are installed by default so, for example, projects using only JS or TS will end up installing many unneeded angular dependencies before this change.

The user experience is also improved as users do not need to manually evaluate which dependencies are required by their projects. Fixes #41.

# Implementation

The repo creates three new public packages: 
- `@ni/eslint-config-javascript`
- `@ni/eslint-config-typescript`
- `@ni/eslint-config-angular`

The previous `@ni/eslint-config` package will be marked deprecated with a message redirecting users to the github for new packages. The new packages' major versions will start at (current package major version +1) to avoid confusion.

The smoke tests are each migrated to their own private packages and more faithfully represent a deployed configuration.

# User impact

One major change is that due to the behavior change of peerDependencies in npm 7, the documentation now describes a simplified workflow for managing the lint packages. Users on npm 7 only need to install their specific package, for example `@ni/eslint-config-javascript` and all peerDependencies are guaranteed to be installed.

So the recommendation when using npm 7 will be to only include the exact package you need and rely on npm cli and the generated lock file for the project to handle the dependency tree. Users will not be encouraged to manually add the peerDependencies of the package they install to their own package.json dependencies (unless they are using npm 6 or below).

This isn't common practice yet but I expect the pattern to be more pervasive as npm 7 usage increases.


CONTRIBUTING.md
```diff
@@ -27,6 +27,7 @@ The rules build on several exisiting rulesets such as the airbnb ruleset, the es
 
 When the command is run it will print a JSON representation of each of the maintained rulesets. In addition the command will print the result of some audits that try to detect patterns we want to avoid, such as rules configured with a `warn` severity level.
 
+<!--
```
rajsite, line 30: I think we should migrate to using [beachball](https://microsoft.github.io/beachball/) for handling the release workflow as we have done in the design system. Has added benefits of enabling good workflows for CHANGLOGS and for avoiding manual tag publish steps.

I'll create an issue to implement the workflow and to update these commented out docs. For now the release workflow will be pretty manual requiring manually bumping versions across all the packages and manually publishing to npm. I imagine that will only be necessary for the first publish (as manual steps are needed to publish a new package anyway) and follow-up publishes will be with beachball integrated.

README.md
```diff
@@ -1,65 +1,89 @@
 <div align="center">
     <div>
-        <!-- Load images from raw.githubusercontent.com to enable image rendering when viewed from https://www.npmjs.com/package/@ni/eslint-config -->
-        <img src="https://raw.githubusercontent.com/ni/javascript-styleguide/HEAD/resources/logo.svg" alt="JavaScript, TypeScript, and NI logo" width="300" height="100">
+        <!-- Load images from raw.githubusercontent.com to enable image rendering when viewed from npmjs.com -->
+        <img src="https://raw.githubusercontent.com/ni/javascript-styleguide/HEAD/docs/logo.svg" alt="JavaScript, TypeScript, and NI logo" width="300" height="100">
```
rajsite, line 4: I moved the docs assets to the docs folder so these won't render until after the package is published and the assets are available at those URLs.

README.md
```diff
@@ -1,65 +1,89 @@
...
     </div>
 </div>
 
 # NI JavaScript and TypeScript Style Guide
 
-[![NPM Version](https://img.shields.io/npm/v/@ni/eslint-config.svg)](https://www.npmjs.com/package/@ni/eslint-config)
+[![NPM Version](https://img.shields.io/npm/v/@ni/eslint-config-javascript.svg)](https://www.npmjs.com/package/@ni/eslint-config-javascript)
```
rajsite, line 10: These shields will not be invalid until the monorepo packages are published.

package.json
```diff
@@ -22,45 +19,23 @@
     "eslintconfig",
     "ni"
   ],
-  "author": "NI",
+  "author": "National Instruments",
```
rajsite, line 22: Legal name / copyright name should stay National Instruments. Marketing and copy uses NI.

README.md
```diff
@@ -1,65 +1,89 @@
...
+        '@ni/eslint-config-typescript/requiring-type-checking'
     ],
     parserOptions: {
         project: 'tsconfig.json'
     }
 }
 ```
 
-### Angular
+### Angular configuration
```
rajsite, line 86: @TrevorKarjanis if you can in particular review the Angular section, I tried to capture the updates to the NI provided configs like we discussed.

packages/eslint-config-angular/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
rajsite, line 1: I'll comment here since it's the first in the file view. All the public packages should behave as if they are standalone and as such include README, LICENSE, and their own standalone configs, ie .eslintrc.js configs. There shouldn't be coupling to other packages without going through standard npm dependency resolution.

tests/print-evaluated-rules/index.js
```diff
@@ -1,19 +1,19 @@
...
         return JSON.stringify(config).indexOf('"warn"') !== -1;
     };
 
     const eslint = new ESLint();
-    const configEslint = await eslint.calculateConfigForFile(`${__dirname}/../test/eslint/index.js`);
-    const configTypescript = await eslint.calculateConfigForFile(`${__dirname}/../test/typescript/index.js`);
-    const configTypescriptTypechecked = await eslint.calculateConfigForFile(`${__dirname}/../test/typescript-type-checking/index.js`);
-    const configAngular = await eslint.calculateConfigForFile(`${__dirname}/../test/angular/index.ts`);
-    const configAngularTemplate = await eslint.calculateConfigForFile(`${__dirname}/../test/angular/index.html`);
+    const configEslint = await eslint.calculateConfigForFile(`${__dirname}/../javascript/index.js`);
```
rajsite, line 12: This package is a known exception to the package dependency rule as it is a meta package comparing implementation details across packages. It is private and should not be published or duplicated. That will be captured in a README for the package.

.github/workflows/build.yml
```diff
@@ -21,23 +21,14 @@ jobs:
     - run: npm ci
 
     # Build
-    - run: npm pack
+    - run: npm run pack
```
rajsite, line 24: This comment is for a top-level comment to discuss concerns / questions about the monorepo approach. Threads better than top-level comments in PR @TrevorKarjanis @jattasNI @mure @LukasKall

tests/print-evaluated-rules/index.js
rajsite: Added a README for the package.

tests/angular/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
jattasNI, line 1: I think it would be good to have a bit of documentation explaining the pattern of having tests in separate packages since I think it's more common to expect tests next to the code. Some ideas:
1. fill in the `description` field for each test package to say something like "This is an unpublished package which contains the tests for ..."
2. Add a README for each test package saying the same thing
3. Add a README at the root of the tests folder describing the pattern/rationale
4. Add a note in the CONTRIBUTING doc explaining the testing strategy

I don't think we need all of these but a few sign posts would help.

LukasKall (approved):

mure (approved):
I looked over the changes and generally agree with the approach. If I have more time later today I'll review in more detail, but I'll probably be in meetings until 12.

README.md
```diff
@@ -1,65 +1,89 @@
...
+    Then follow the [TypeScript configuration](#typescript-configuration) instructions.
+- For Angular run:
+
+    ```bash
+    npm install @ni/eslint-config-angular
+    ```
+
+    Then follow the [Angular configuration](#angular-configuration) instructions.
+
+If you are using npm 6 or lower, use the following instructions to manually install peer dependencies:
```
jattasNI, line 42: We could be a bit more explicit that the npm 6 workflow also includes the steps above to pick the right package and to follow the language-specific configuration instructions. Right now it reads like those steps are only needed for Angular 7. Maybe reframe the npm 6 steps as an additional step in between running the install command and following the configuration instructions?

packages/eslint-config-angular/.eslintrc.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-    extends: './index',
+    extends: '@ni/eslint-config-javascript',
```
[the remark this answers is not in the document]
jattasNI: Oh, never mind, this is the config file for linting this package, not the config file for consumers of this package like I thought (that's in `index.js`).

jattasNI (approved):

TrevorKarjanis (approved):

rajsite:
Failed to bowl and stroll :P I can pick-it up after returning. Alternatively if someone wants to merge it they have to do the work to create new packages under the ni org and assign them permissions etc before they can be published. Can work with Fred or Mark if feeling ambitious.

We may want a minor push to main of the Angular README updates in-case someone does try and integrate with the exisiting package before the monorepo transition. @TrevorKarjanis

tests/angular/package.json
rajsite: Added docs to Contributing.md, READMEs pointing to the docs, and description fields to the package.json files

README.md
rajsite: Clarified that all projects need to install a lint package but the manual peer dependencies install is an additional step specific to npm 6 and below

PR #61, opened 2021-09-09
Add docs for unbound-method

jattasNI:
# Rationale

Fixing violations of the [@typescript-eslint/unbound-method](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/unbound-method.md) rule requires you to understand how `this` behaves in various situations, which is something I always have to re-teach myself. Rather than sending developers out to the web to learn how to fix violations, we should provide some guidance in our rule documentation.

# Implementation

Synthesized information from [this comment thread](https://github.com/ni/javascript-styleguide/pull/20/files#r595574910) when we initially discussed the rule and [this stackblitz](https://stackblitz.com/edit/typescript-xdrtzi) where I re-learned `this` behaviors into a comment which describes some common resolutions.

### DIFF
--- a/typescript-requiring-type-checking.js
+++ b/typescript-requiring-type-checking.js
@@ -50,7 +50,7 @@ module.exports = {
                     // Bad, removeEventListener isn't removing the original handler
                     elt.addEventListener('click', event => this.clickHandler());
                     elt.removeEventListener('click', event => this.clickHandler());
-            3. Changing the function to be static if it doesn't need to refer to the `this` instance.
+            3. If the function doesn't use `this` to refer to an instance or other static members, you can change the function to be static and the rule will ignore it.
         */
         '@typescript-eslint/unbound-method': ['error', { ignoreStatic: true }],
 
--- a/typescript-requiring-type-checking.js
+++ b/typescript-requiring-type-checking.js
@@ -50,6 +50,7 @@ module.exports = {
                     // Bad, removeEventListener isn't removing the original handler
                     elt.addEventListener('click', event => this.clickHandler());
                     elt.removeEventListener('click', event => this.clickHandler());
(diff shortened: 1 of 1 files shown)


LukasKall (approved):

typescript-requiring-type-checking.js
```diff
@@ -21,6 +21,37 @@ module.exports = {
...
+                    elt.removeEventListener('click', handler);
+
+                    // Bad, violates this rule
+                    elt.addEventListener('click', this.clickHandler);
+                    elt.removeEventListener('click', this.clickHandler);
+
+                    // Bad, removeEventListener isn't removing the original handler
+                    elt.addEventListener('click', event => this.clickHandler());
+                    elt.removeEventListener('click', event => this.clickHandler());
+            3. Changing the function to be static if it doesn't need to refer to the `this` instance.
```
TrevorKarjanis, line 53: I don't think this fixes the issue, because static methods can just as well use `this` to refer to the class and call other static methods.

TrevorKarjanis (changes requested):

typescript-requiring-type-checking.js
jattasNI: True. My understanding is that `ignoreStatic` will cause the rule not to fire for static methods, which would be correct if the static method doesn't use `this`. I tried re-wording it to express this; let me know what you think.

TrevorKarjanis: I guess with respect to how the rule is configured that is correct. However, there is nothing enforcing static methods not use `this`. Someone could come along later and one. Maybe that is a better discussion around the rule's configuration rather than what your intentions are with this PR.

TrevorKarjanis (approved):

PR #72, opened 2021-11-16
Add noUselessIndex to no-useless-path-segments

rajsite:
# Justification

Found code that was mixing usages of index imports and directory imports.

For example with the following npm package file structure:
```
- src
  - awesome
    - index.ts
  - other.ts
```

Some imports within the library would use `import './awesome';` while others used `import './awesome/index';`.

In es6 modules world those resolve to two different absolute paths, ie `C:/lib/src/awesome` vs `C:/lib/src/awesome/index` so tooling that may not be aware of commonjs file resolution [may treat them as different symbols](https://github.com/psykolm22/angular-google-place/issues/29#issuecomment-360884843). This is not a concern in modern versions of tooling that I have seen but enforcing the consistency seems useful anyway for readability.

This PR copies the exisiting airbnb rule for [import/no-useless-path-segments](https://github.com/airbnb/javascript/blob/d8cb404da74c302506f91e5928f30cc75109e74d/packages/eslint-config-airbnb-base/rules/imports.js#L241) and adds the [noUselessIndex](https://github.com/import-js/eslint-plugin-import/blob/7c239fed485ea0785a96c1fa2045d96c181bb79c/docs/rules/no-useless-path-segments.md#nouselessindex) configuration.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7624,7 +7624,7 @@
         "@angular/core": "^12.0.0"
       },
       "devDependencies": {
-        "@ni/eslint-config-angular": "^3.3.0",
+        "@ni/eslint-config-angular": "*",
         "@types/jasminewd2": "^2.0.10"
       }
     },
@@ -7632,7 +7632,7 @@
       "name": "@ni/javascript-test",
       "version": "1.0.0",
       "devDependencies": {
-        "@ni/eslint-config-javascript": "^3.0.3"
+        "@ni/eslint-config-javascript": "*"
       }
     },
     "tests/print-evaluated-rules": {
@@ -7641,7 +7641,7 @@
       "dependencies": {
         "@angular-eslint/eslint-plugin": "*",
         "@ni/angular-test": "*",
-        "@ni/eslint-config-angular": "^3.3.0",
+        "@ni/eslint-config-angular": "*",
         "@ni/javascript-test": "*",
         "@ni/typescript-requiring-type-checking-test": "*",
         "@ni/typescript-test": "*",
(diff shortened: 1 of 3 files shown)


LukasKall (approved):

package-lock.json
```diff
@@ -127,57 +127,60 @@
       "peer": true
     },
     "node_modules/@angular-eslint/builder": {
-      "version": "12.5.0",
-      "resolved": "https://registry.npmjs.org/@angular-eslint/builder/-/builder-12.5.0.tgz",
-      "integrity": "sha512-abex1KdqEvUYbJqLkqhEeW4r0DuJPDzlMZYODMyBDDFTcBFYQCzLFUkhIjRNet9Vjto3Y3DbkkFUfS4DEEwjTw==",
+      "version": "12.6.1",
```
jattasNI, line 130: Are the changes to `package-lock.json` intentional?

jattasNI (approved):

package-lock.json
rajsite: yea npm had to make those updates to resolve peerDependency versions. I didn't try to dig into why but I did git clean the directory and try again and it was still needed.

packages/eslint-config-javascript/index.js
```diff
@@ -74,6 +74,18 @@ module.exports = {
         */
         'import/no-default-export': 'error',
 
+        /*
+            Keep path names short and consistent by avoiding unnecessary realtive path segments and consistently
```
TrevorKarjanis, line 78: Misspelled relative

TrevorKarjanis (approved):

mure (approved):

PR #75, opened 2021-12-13
Disable @angular-eslint/component-max-inline-declarations in tests

jattasNI:
The `@angular-eslint/component-max-inline-declarations` rule enforces that inline component templates (those written in the .ts file) can only contain a 15 lines before they should be moved to a separate file. I propose this rule should be disabled in test files since it improves test readability to have helper components in the same file as the test, even if they are longer than 15 lines. We already disable a couple related rules with similar justification (see below in the file I'm editing).

This was spurred by [this PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/243777?_a=files&path=/Web/Workspaces/SystemLinkShared/projects/systemlink-lib-angular/src/components/sl-toolbar/sl-toolbar.component.spec.ts) where someone disabled this rule in a test file. In this case the repo is using an older version of our ruleset so the rule was firing on smaller components, but I believe suppressing it would still be valid even if the component were larger.

### DIFF
--- a/packages/eslint-config-angular/index.js
+++ b/packages/eslint-config-angular/index.js
@@ -112,6 +112,12 @@ module.exports = {
             files: ['*.spec.ts'],
             rules: {
                 /*
+                    Tests often define helper components and it improves test readability if they are in the same
+                    file as the tests.
+                */
+                '@angular-eslint/component-max-inline-declarations': 'off',
+
+                /*
                     Components defined in tests are typically only used within a single test file so don't need.
                     a compontent selector.
                 */
--- a/change/@ni-eslint-config-angular-d7b7362b-1d9a-4a43-8b03-f444ec709ea4.json
+++ b/change/@ni-eslint-config-angular-d7b7362b-1d9a-4a43-8b03-f444ec709ea4.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Disable @angular-eslint/component-max-inline-declarations in tests",
+  "packageName": "@ni/eslint-config-angular",
+  "email": "jattasNI@users.noreply.github.com",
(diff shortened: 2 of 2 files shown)


packages/eslint-config-angular/index.js
```diff
@@ -111,6 +111,12 @@ module.exports = {
         {
             files: ['*.spec.ts'],
             rules: {
+                /*
+                    Tests often define helper components and it improves test readability if they are in the same
+                    file as the tests.
+                */
+                '@angular-eslint/component-max-inline-declarations': 'off',
```
TrevorKarjanis, line 118: I'm alright with this change. However, I'm surprised that tests need more than 15 lines for host mocks. It sounds like the PR you mentioned would have been resolved with 15? Did we consider this?

rajsite (approved):

packages/eslint-config-angular/index.js
jattasNI: Correct, that PR wouldn't have needed to suppress the rule if they were using a more recent version of the ruleset. I'd also be surprised to see such large components in a test, but I posit that we'd still prefer those large components to live in the test file and not a separate file, so I'm updating the rule to match that philosophy.

TrevorKarjanis: Alright, cool.

TrevorKarjanis (approved):

LukasKall (approved):

jattasNI:
@mure This PR is awaiting your vote!

jattasNI:
@rajsite, @mure is on vacation for the rest of the year. This rule seems fairly noncontroversial so I think it would be ok to merge without all 4 votes.

PR #77, opened 2022-01-07
Beachball integration retroactive PR

rajsite:
# Pull Request

## 🤨 Rationale

Retroactive PR for beachball workflow publish. Uses beachball to automate package version bumps, saving version bumps in git, and publishing bumped packages to npm.

Note: PR is to collect feedback on commits pushed to main. Feedback will be addressed in a follow-up PR.

## 👩‍💻 Implementation

- Copied the beachball.config.js tool for bumping package-lock.json versions from [nimble](https://github.com/ni/nimble/blob/d5ed889af8947b8a0ec94170258f6cab14fdbffb/beachball.config.js).
- Bumped beachball to version including the postbump hook.
- Changed the test package dependency versions to be '*' so they don't need to be bumped on package updates (they are private so no benefit bumping versions in them).

## 🧪 Testing

Verified publish from merging #75 which bumped versions in 9835383148f958aac4187829c9ea38afb2dcd673

## ✅ Checklist

- [x] I have updated the project documentation to reflect my changes or determined no changes are needed.

### DIFF
--- a/beachball.config.js
+++ b/beachball.config.js
@@ -1,0 +1,73 @@
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+
+// Workaround for the following issues:
+// Beachball issue: https://github.com/microsoft/beachball/issues/525
+// NPM issue 1: https://github.com/npm/cli/issues/3756
+// NPM issue 2: https://github.com/npm/cli/issues/3757
+
+/**
+ * Returns a beachball postbump hook implementation that bumps package
+ * references specified in the root package-lock.json file during
+ * beachball package version bumps.
+ */
+const postbump = (_packagePath, packageName, packageVersion) => {
+    console.log(`Updating lockfile for package ${packageName} to version ${packageVersion}`);
+    const lockPath = path.resolve(__dirname, './package-lock.json');
+
+    console.log(`Lockfile to update: ${lockPath}`);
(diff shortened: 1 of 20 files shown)


beachball.config.js
```diff
@@ -0,0 +1,73 @@
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+
```
jattasNI, line 4: Since this is duplicated from Nimble it could be worth adding a comment to both copies that links them to each other to remind us to keep them in sync if one of them changes.

jattasNI (approved):
Thanks for taking this on! I'm excited to get to use beachball's tooling with actual semver instead of always using "prerelease" like we do in Nimble right now.

beachball.config.js
rajsite: Maybe we should actually publish this as a package... I can see it being useful for other projects. For example I'd like to integrate it in Vireo as well.

jattasNI: Even better!

TrevorKarjanis (approved):
I've got almost 10000 views of the [beach near where I'm at](https://goo.gl/maps/vWwLSS4NvtATMSE2A).

beachball.config.js
rajsite: That package is in-work, I'll close this out for now and expect to integrate that package when ready: https://github.com/ni/nimble/pull/248

PR #82, opened 2022-01-16
Upgrade production dependencies with Angular 14 support.

TrevorKarjanis:
- Upgrade eslint, @typescript-eslint, @angular-eslint, and eslint-config-airbnb-base major versions.
- Add a diff script to compare old and new rule sets.
- Fix a bug caused by incorrect `devDependencies` in `@ni/eslint-config-typescript`.
- Fix the character casing of npm and ESLint.

This PR tries not to break downstream consumers by introducing changes to the rule set.
- The diffs of the rule set after upgrading and after addressing breaking changes are attached.
  - [upgrade-rules-diff.txt](https://github.com/ni/javascript-styleguide/files/9338362/upgrade-rules-diff.txt)
  - [fix-rules-diff.txt](https://github.com/ni/javascript-styleguide/files/9338361/fix-rules-diff.txt)
- import/no-extraneous-dependencies has a complex configuration and only added "**/.eslintrc.js" to its devDependencies option which I don't think is a breaking change.
- no-confusing-arrow added the onlyOneSimpleParam option, but uses the default value.
- @angular-eslint/template/i18n has a complex configuration and enabled the checkDuplicateId option, but NI doesn't specify IDs.
- Changes to the implementation of already enabled rules can cause errors.

Notes
- This is required to support the latest ECMA features like private fields.
- This is required to support Angular 14.
- Drop support for Angular 11 and below.
- The `@angular-eslint/schematics` dependency [ensures ESLint is configured](https://github.com/angular-eslint/angular-eslint#using-eslint-by-default-when-generating-new-projects-within-your-workspace) for newly generated projects.
- As per their branding, npm should never be uppercase.

Linted and tested
- Manually verified that upgrading the package in a consumer works correctly
- Verified that automated testing passes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ node_modules
 *.tgz
 .vscode
 .npmrc
+./rules.txt
+./rules-diff.txt
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,13 +19,15 @@ When updating dependencies the `npm run print-available-rules` command can be us
 
 **Note:** This command does not print what rules are currently used but instead prints all the rules that are available.
 
-The TypeScript extension rules (which are rules that extend the behavior of exisiting ESlint rules) must be used in place of corresponding ESlint rules. Printing the available rules is useful for manually auditing what extension rules need to be used.
+The TypeScript extension rules (which are rules that extend the behavior of existing ESlint rules) must be used in place of corresponding ESlint rules. Printing the available rules is useful for manually auditing what extension rules need to be used.
 
 ### Checking the evaluated rules
 
-The rules build on several existing rulesets such as the airbnb ruleset, the ESlint recommended ruleset, the TypeScript recommended ruleset, etc. The `npm run print-evaluated-rules` command helps to audit the final resolved rule values.
+The rules build on several existing rulesets such as the airbnb ruleset, the ESlint recommended ruleset, the TypeScript recommended ruleset, etc. The `npm run print-evaluated-rules` command helps to audit the final resolved rule values. It will print the result of some audits that try to detect patterns we want to avoid, such as rules configured with a `warn` severity level.
 
-When the command is run it will print the result of some audits that try to detect patterns we want to avoid, such as rules configured with a `warn` severity level.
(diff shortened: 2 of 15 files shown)


TrevorKarjanis:
> @TrevorKarjanis is there anything blocking us from taking this PR out of Draft state and getting it reviewed/submitted?

The biggest pain here was identifying new rules. I did so manually at first. My next step is to use the utility Milan created to print a before and after, and compare them to verify the findings described here. We then need to decide on our policy. My thought is to disable them if they are recommended, and create an issue to evaluate all of them appropriately so that upgrade PRs aren't blocked by a potentially lengthy process. As we saw in our initial evaluation, blindly enabling them is not practical. We could also take a quick pass to identify obvious cases that align or violate existing conventions.

After that, the dependencies in the PR just need to be upgraded again. Run the tests, update the README, and post the PR.

jattasNI:
Let me know if I can help in any way. This iteration our team is driving the Angular 13 upgrade in many of the apps and libs that depend on this, so getting this in will be an important step and something we'd be happy to help with.

rbell517:
Is there an update on this PR? I'm trying to apply these rules to Skyline's EndToEndTests subdir and its using ES2022 features. I can change the handful of places we use those features but if this is close then I might as well wait.

TrevorKarjanis:
> Is there an update on this PR? I'm trying to apply these rules to Skyline's EndToEndTests subdir and its using ES2022 features. I can change the handful of places we use those features but if this is close then I might as well wait.

You're adding linting the the end-to-end tests? You're a hero! This effort will hopefully restart next week, as myself and many involved are heavily involved in the latest customer effort. This is also a part of our larger migration to Angular 13. I would suggest regressing those features temporarily as that is likely much easier than this.

rajsite:
@TrevorKarjanis FYI you will need to create a new PR based on a branch from the ni repo instead of branch from your fork to merge

TrevorKarjanis:
> @TrevorKarjanis FYI you will need to create a new PR based on a branch from the ni repo instead of branch from your fork to merge

I happen to be doing exactly that! Thanks.

PR #104, opened 2022-09-02
Upgrade production dependencies with Angular 13 support.

TrevorKarjanis:
- Upgrade eslint, @typescript-eslint, @angular-eslint, and eslint-config-airbnb-base with support for Angular 14.
- Add a diff script to compare old and new rule sets.
- Fix a bug caused by incorrect `devDependencies` in `@ni/eslint-config-typescript`.
- Fix the character casing of npm and ESLint.

This PR tries not to break downstream consumers by introducing changes to the rule set.
- The diffs of the rule set after upgrading and after addressing breaking changes are attached.
  - [upgrade-rules-diff.txt](https://github.com/ni/javascript-styleguide/files/9338362/upgrade-rules-diff.txt)
  - [fix-rules-diff.txt](https://github.com/ni/javascript-styleguide/files/9338361/fix-rules-diff.txt)
- import/no-extraneous-dependencies has a complex configuration and only added "**/.eslintrc.js" to its devDependencies option which I don't think is a breaking change.
- no-confusing-arrow added the onlyOneSimpleParam option, but uses the default value.
- @angular-eslint/template/i18n has a complex configuration and enabled the checkDuplicateId option, but NI doesn't specify IDs.
- Changes to the implementation of already enabled rules can cause errors.

Notes
- This is required to support the latest ECMA features like private fields.
- This is required to support Angular 13.
- Drop support for Angular 11 and below.
- The `@angular-eslint/schematics` dependency [ensures ESLint is configured](https://github.com/angular-eslint/angular-eslint#using-eslint-by-default-when-generating-new-projects-within-your-workspace) for newly generated projects.
- As per their branding, npm should never be uppercase.

Linted and tested
- Verified that automated testing passes


rajsite (approved):

packages/eslint-config-javascript/index.js
```diff
@@ -358,6 +358,26 @@ module.exports = {
             anonymous: 'always',
             named: 'never',
             asyncArrow: 'always'
-        }]
+        }],
+
+        /*
+            The following are rules that would be enabled or the configuration changed as part of the upgrade to ESLint
```
jattasNI, line 364: Could you file an issue reminding us to evaluate these and enable the ones we agree with?

TrevorKarjanis: #112

PR #113, opened 2022-11-11
Loosen @typescript-eslint/naming-convention to allow properties with custom-element-names

### LINKED ISSUE #111: Don't enforce naming convention for HTMLElementNameTagMap
Defining an `HTMLElementTagNameMap` interface [is common in TypeScript](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.htmlelementtagnamemap.html) libraries that define elements so that `querySelector` can return a typed element when you select by the element name. With our current ruleset you have to suppress `@typescript-eslint/naming-convention` because the interface member must match the element name which is hyphen-case instead of camelCase.
```ts
declare global {
    interface HTMLElementTagNameMap {
        // eslint-disable-next-line @typescript-eslint/naming-convention
        'sl-action-menu': NgElement & WithProperties<SlActionMenuComponent>;
    }
}
```

We can probably improve this by using one of the strategies described in [Ignore properties that require quotes](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/naming-convention.md#ignore-properties-that-require-quotes)

This came from [this PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/354794?_a=files&path=/Web/Workspaces/SystemLinkShared/projects/file-manager-lib) we are updating a repo with the latest version of the styleguide and had to add a bunch of suppressions.

jattasNI:
# Justification

Fixes #111.

# Implementation

Based off the third example in the rule's documentation under [Ignore properties that require quotes](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/naming-convention.md#ignore-properties-that-require-quotes), I am using a regex to match properties that follow the lower-hyphen-case naming convention of custom elements, and turning off format checks for properties that match.

# Testing

Locally tried these cases with the results described in the comments.

```ts
declare global {
    interface HTMLElementTagNameMap {
        'valid-custom-element': NI; // passes because of new configuration
        'valid-custom-element2': NI; // passes because of new configuration
        '2invalid-custom-element': NI; // fails because custom-element regex requires starting with a-z

        'nohyphen': NI; // passes because camelCase (with no capitals) is allowed by original configuration
        'noHyphen': NI; // passes because camelCase is allowed by original configuration
        camelCase: NI; // passes because camelCase is allowed by original configuration

        'Capital-Custom-Element': NI; // fails because PascalCase is disallowed by original configuration and new regex only matches lower case
        'capital-Custom-Element': NI; // fails because PascalCase is disallowed by original configuration and new regex only matches lower case
        PascalCase: NI; // fails because PascalCase is disallowed by original configuration
    }
}
```

Also did some testing in nimble and SystemLinkShared with locally built packages. This allows us to get rid of a global suppression in nimble and only add one inline suppression. It doesn't really help with SystemLinkShared because the one place that uses this pattern is an older library that has a custom `naming-convention` configuration so the one from the styleguide doesn't apply. However we should be able to copy the same pattern from this PR into that custom configuration.

### DIFF
--- a/packages/eslint-config-typescript/requiring-type-checking.js
+++ b/packages/eslint-config-typescript/requiring-type-checking.js
@@ -53,13 +53,23 @@ module.exports = {
         */
 
         /*
-            The default configuration for this rule requires that enum members
-            follow the camelCase style, which doesn't match the existing convention
-            within NI and the broader TS community to use PascalCase. The rest of the
-            configuration is identitical to the default.
+            Most of this configuration is identitical to the default with exceptions called out in comments below
         */
         '@typescript-eslint/naming-convention': [
             'error',
+            // Allow properties with custom-element-names, commonly used in HTMLElementNameTagMap
+            {
+                selector: 'property',
+                format: null,
+                filter: {
+                    // custom element regex from https://html-validate.org/rules/element-name.html
+                    regex: '[a-z][a-z0-9\\-._]*-[a-z0-9\\-._]*$',
+                    match: true
+                }
+            },
+            // The default configuration for this rule requires that enum members
+            // follow the camelCase style, which doesn't match the existing convention
+            // within NI and the broader TS community to use PascalCase.
             {
                 selector: 'enumMember',
                 format: ['PascalCase'],
--- a/packages/eslint-config-typescript/requiring-type-checking.js
+++ b/packages/eslint-config-typescript/requiring-type-checking.js
@@ -62,8 +62,8 @@ module.exports = {
                 selector: 'property',
                 format: null,
                 filter: {
-                    // custom element regex from https://html-validate.org/rules/element-name.html
-                    regex: '[a-z][a-z0-9\\-._]*-[a-z0-9\\-._]*$',
+                    // custom element regex adapted from https://html-validate.org/rules/element-name.html
+                    regex: '^[a-z][a-z0-9\\-._]*-[a-z0-9\\-._]*$',
                     match: true
                 }
             },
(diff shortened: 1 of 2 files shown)


packages/eslint-config-typescript/requiring-type-checking.js
```diff
@@ -53,13 +53,23 @@ module.exports = {
...
         */
         '@typescript-eslint/naming-convention': [
             'error',
+            // Allow properties with custom-element-names, commonly used in HTMLElementNameTagMap
+            {
+                selector: 'property',
+                format: null,
+                filter: {
+                    // custom element regex from https://html-validate.org/rules/element-name.html
+                    regex: '[a-z][a-z0-9\\-._]*-[a-z0-9\\-._]*$',
```
TrevorKarjanis, line 66: Note, this requires a dash (-). Technically, I don't think that is required.

TrevorKarjanis (approved):

packages/eslint-config-typescript/requiring-type-checking.js
rajsite: A dash is required for `custom elements` as the browser feature (names without `-` are reserved for HTML). Are there use cases to increase the scope outside of custom elements (don't think so, just checking if there was something on your mind)?

TrevorKarjanis: No, I wasn't aware of that requirement. Thanks.

rajsite, line 66: > All are correct except for 2invalid-custom-element

I think the reason for that is the regular expression does not match to the start of the string (missing `^` at the start) but is just matching to the end of the string (it does match to `$`)

So maybe the following would resolve that?

```suggestion
                    regex: '^[a-z][a-z0-9\\-._]*-[a-z0-9\\-._]*$',
```

It also depends on what string is actually being matched against. If it's the full line of source that would not work. Other stuff on the line would interfere.

If it is just the string name token, ie `2invalid-curom-element` that would fix it. My guess it is just the name token so that might fix it.

rajsite (approved):

packages/eslint-config-typescript/requiring-type-checking.js
jattasNI: Yep, that was it, thanks! I made that change and updated the PR description.

PR #120, opened 2023-01-26
New Playwright ESLint package

### LINKED ISSUE #114: NI ESLint ruleset for Playwright
Several codebases are using Playwright for end-to-end testing. They often depend on [`eslint-plugin-playwright`](https://www.npmjs.com/package/eslint-plugin-playwright) and configure their `eslintrc` like this:
```
           "extends": [
                "@ni/eslint-config-typescript",
                "@ni/eslint-config-typescript/requiring-type-checking",
                "plugin:playwright/playwright-test"
            ],
```

One problem with this configuration is that `playwright/playwright-test` has several rules configured as `warning` but we generally prefer to use `error` or `off` because it prevents lint output from getting noisy with warnings that devs don't bother to fix.

If we publish a Playwright-specific configuration we should also:
1. have it extend `@ni/eslint-config-typescript` so clients don't need to extend as many rulesets. Having it in a separate config from our typescript ruleset would allow it to be targeted to specific files with the `-test.ts` extension. We'd also provide a `eslint-config-playwright-requiring-type-checking` that would be an alias for the existing typescript-requring-type-checking (or decide to get rid of all of our `*-requring-type-checking` rulesets, but that could be done in a separate pass).
2. evaluate rules that aren't part of their recommended ruleset to see if we want to enable them by default (see comments below for our decisions)
3. have a smoke test as we do for other config files
4. update our scripts that evaluate whether the rules have changed
5. update docs
6. evangelize the ruleset to playwright codebases (and/or update [those codebases](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=%40playwright/test%20file%3Apackage.json&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//Web/Workspaces/RoutineManager/acceptance-tests/package.json) ourselves)

jattasNI:
# Justification

Fixes #114.

# Implementation

1. Added a new package containing playwright rulesets
    - it exposes regular and requiring-type-checking variants which derive from the base typescript ruleset
    - it adds configurations for rules as discussed in #114 
    - updated root README to mention it and describe its configuration
2. Added smoke tests for the package. I initialized these using Playwright's setup instructions then deleted a bunch of unnecessary stuff since we don't actually execute the Playwright test.
3. I noticed that the `npm run lint` command was failing on existing code so I fixed violations (just whitespace) and enabled it in the pipeline.


tests/playwright-requiring-type-checking/playwright.config.ts
```diff
@@ -0,0 +1,78 @@
+import { defineConfig, devices } from '@playwright/test';
```
rajsite, line 1: we nuked all of angular's config (angular.json, etc) so if this is also unused may as well nuke it too

tests/playwright-requiring-type-checking/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+    "files": ["./index.ts", "./playwright.config.ts"],
+    "include": ["../"],
```
rajsite, line 3: I don't think this includes is needed

tests/playwright/playwright.config.ts
```diff
@@ -0,0 +1,78 @@
+import { defineConfig, devices } from '@playwright/test';
```
rajsite, line 1: yea nuke it!

rajsite (approved):

packages/eslint-config-playwright/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "/*.js",
+    "!/.*.js"
+  ],
+  "dependencies": {
+    "@ni/eslint-config-typescript": "^4.2.0",
+    "eslint-plugin-playwright": "^0.12.0"
```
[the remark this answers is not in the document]
rajsite: (fyi make sure to re-run npm install to update the lock file after)

TrevorKarjanis (approved):

PR #129, opened 2023-11-02
Add support for Angular 15

mollykreis:
- Update the Angular tests to run against Angular 15 rather than Angular 14
- Update the peer dependencies of `eslint-config-angular` to include Angular 15 in addition to Angular 14

### DIFF
--- a/change/@ni-eslint-config-angular-ebf2bbaa-353f-4c86-81c2-7956d462fc41.json
+++ b/change/@ni-eslint-config-angular-ebf2bbaa-353f-4c86-81c2-7956d462fc41.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add support for Angular 15",
+  "packageName": "@ni/eslint-config-angular",
+  "email": "20542556+mollykreis@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/eslint-config-angular/package.json
+++ b/packages/eslint-config-angular/package.json
@@ -31,11 +31,11 @@
     "!/.*.js"
   ],
   "peerDependencies": {
-    "@angular-eslint/builder": "^15.2.0",
-    "@angular-eslint/eslint-plugin": "^15.2.0",
-    "@angular-eslint/eslint-plugin-template": "^15.2.0",
-    "@angular-eslint/schematics": "^15.2.0",
-    "@angular-eslint/template-parser": "^15.2.0",
+    "@angular-eslint/builder": "^14.0.2 || ^15.2.0",
+    "@angular-eslint/eslint-plugin": "^14.0.2 || ^15.2.0",
+    "@angular-eslint/eslint-plugin-template": "^14.0.2 || ^15.2.0",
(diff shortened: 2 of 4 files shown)


TrevorKarjanis (approved):

packages/eslint-config-angular/package.json
```diff
@@ -31,11 +31,11 @@
     "!/.*.js"
   ],
   "peerDependencies": {
-    "@angular-eslint/builder": "^14.0.2",
-    "@angular-eslint/eslint-plugin": "^14.0.2",
-    "@angular-eslint/eslint-plugin-template": "^14.0.2",
-    "@angular-eslint/schematics": "^14.0.2",
-    "@angular-eslint/template-parser": "^14.0.2",
+    "@angular-eslint/builder": "^14.0.2 || ^15.2.0",
```
jattasNI, line 34: I think I'd vote to change these to `"14.0.0 || 15.0.0"`. That should allow clients to use any minor or patch version of 14 or 15 rather than just versions greater than 14.0.2 or 15.2. We'd still leave the other package file at a specific higher version for testing within this repo.

It's unlikely to make a difference for Skyline clients who will start at 15.2 but it better expresses our intent and might help some other client on an earlier version.

You should rerun `npm install` to regenerate `package-lock.json` after making that change. I doubt it will result in any diffs but it's better to be safe.

jattasNI (approved):

packages/eslint-config-angular/package.json
jattasNI: Yes, my bad, It should have the `^`.

change/@ni-eslint-config-angular-ebf2bbaa-353f-4c86-81c2-7956d462fc41.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: Did you sanity check the changes against an existing project (ie with the packages tgz) that is still on 14 to make sure nothing funny happens?

rajsite (approved):

packages/eslint-config-angular/package.json
mollykreis: I've changed this to `^14.0.0 || ^15.0.0`, and I confirmed that could be installed in SystemLinkShared, which is currently using Angular 14.2.

change/@ni-eslint-config-angular-ebf2bbaa-353f-4c86-81c2-7956d462fc41.json
rajsite: sounds liek that was addressed here 👍 https://github.com/ni/javascript-styleguide/pull/129#discussion_r1380445691

rajsite:
@mollykreis fyi the release failed: https://github.com/ni/javascript-styleguide/actions/runs/6735793538/job/18309763573#step:13:38

looking into it

rajsite:
> @mollykreis fyi the release failed: https://github.com/ni/javascript-styleguide/actions/runs/6735793538/job/18309763573#step:13:38
> 
> looking into it

Release is published! 🚢

PR #133, opened 2023-11-13
Update to Angular 15

mollykreis:
Update the style guide to support only Angular 15.

I've verified that this package can be used in a SLE application migrated to Angular 15 without any issues with the npm installation.


rajsite (approved):

rajsite:
FYI @jattasNI @TrevorKarjanis @mure straightforward split of the Angular 15 into a standalone major bump, merging before review but can address concerns in a follow-up if needed.

PR #138, opened 2024-03-18
Update angular-eslint to 16

m-akinc:
- Updated dependency on `@angular/core` to `16.2.12` (latest <17) and `@angular-eslint` packages to `16.3.1` (latest <17)
- As the `@angular-eslint/recommended--extra` config no longer exists, I inlined the configuration that [used to be in that file](https://github.com/angular-eslint/angular-eslint/blob/v15.2.1/packages/eslint-plugin/src/configs/recommended--extra.json) in lieu of extending it.
- In our template config, we now also extend the new `@angular-eslint/template/accessibility` config. This config enables two rules that we were not previously explicitly enabling/disabling: `role-has-required-aria` and `interactive-supports-focus`. These two rules may break clients on upgrade. Also, the accessibility rules referenced by the config have been renamed to drop the `accessibility-` prefix, so I updated the names where we had been disabling those rules.
- Updated all urls in the repo to reference "main" instead of "master"

### DIFF
--- a/change/@ni-eslint-config-angular-7bcf1406-f6fc-4336-a2ff-a2fbb32de903.json
+++ b/change/@ni-eslint-config-angular-7bcf1406-f6fc-4336-a2ff-a2fbb32de903.json
@@ -1,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Update to Angular 16",
+  "packageName": "@ni/eslint-config-angular",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/eslint-config-angular/index.js
+++ b/packages/eslint-config-angular/index.js
@@ -1,14 +1,13 @@
 module.exports = {
     extends: [
         'plugin:@angular-eslint/recommended',
-        'plugin:@angular-eslint/recommended--extra',
         'plugin:@angular-eslint/template/process-inline-templates',
         '@ni/eslint-config-typescript'
     ],
     rules: {
         /*
             Overrides to Angular recommended rules:
-            https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/src/configs/recommended.json
+            https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.json
         */
 
         /*
@@ -29,9 +28,44 @@ module.exports = {
         '@angular-eslint/use-lifecycle-interface': 'error',
 
         /*
-            Overrides to Angular extra recommended rules:
-            https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/src/configs/recommended--extra.json
+            Formerly recommended "extra" Angular rules:
+            https://github.com/angular-eslint/angular-eslint/blob/v15.2.1/packages/eslint-plugin/src/configs/recommended--extra.json
         */
+        'no-restricted-imports': [
+            'error',
+            {
+                paths: [
+                    {
+                        name: 'rxjs/Rx',
+                        message: "Please import directly from 'rxjs' instead"
+                    }
+                ]
+            }
+        ],
+        '@typescript-eslint/member-ordering': [
+            'error',
+            {
+                default: [
+                    'static-field',
+                    'instance-field',
+                    'static-method',
+                    'instance-method'
+                ]
+            }
+        ],
+        'no-restricted-syntax': [
+            'error',
+            {
+                selector: 'CallExpression[callee.object.name="console"][callee.property.name=/^(debug|info|time|timeEnd|trace)$/]',
+                message: 'Unexpected property on console object was called'
+            }
+        ],
+        '@typescript-eslint/no-inferrable-types': [
+            'error',
+            { ignoreParameters: true }
+        ],
+        '@typescript-eslint/no-non-null-assertion': 'error',
+        'no-fallthrough': 'error',
 
         /*
             Overrides to Angular rules outside of the recommended configuration:
--- a/packages/eslint-config-angular/template.js
+++ b/packages/eslint-config-angular/template.js
@@ -1,39 +1,31 @@
 module.exports = {
     extends: [
-        'plugin:@angular-eslint/template/recommended'
+        'plugin:@angular-eslint/template/recommended',
+        'plugin:@angular-eslint/template/accessibility'
     ],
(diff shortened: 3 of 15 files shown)


packages/eslint-config-angular/index.js
```diff
@@ -1,14 +1,13 @@
 module.exports = {
     extends: [
         'plugin:@angular-eslint/recommended',
-        'plugin:@angular-eslint/recommended--extra',
```
TrevorKarjanis, line 4: Do we know why recommended--extra was removed?

TrevorKarjanis (approved):

packages/eslint-config-angular/index.js
m-akinc: No, [the release notes](https://github.com/angular-eslint/angular-eslint/blob/main/CHANGELOG.md#breaking-changes-1) just say it was removed (along with a couple others). Couldn't find a commit or issue that explained further.

packages/eslint-config-angular/template.js
```diff
@@ -1,39 +1,30 @@
...
-        '@angular-eslint/template/accessibility-elements-content': 'off',
-
-        /* [accessibility] */
-        '@angular-eslint/template/accessibility-label-for': 'off',
-
-        /* [accessibility] */
-        '@angular-eslint/template/accessibility-label-has-associated-control': 'off',
-
-        /* [accessibility] */
-        '@angular-eslint/template/accessibility-table-scope': 'off',
```
rajsite, line 28: Why were all the `[accessibility]` markings removed?..
See: https://github.com/ni/javascript-styleguide?tab=readme-ov-file#accessibility

packages/eslint-config-angular/index.js
TrevorKarjanis: Thanks for looking.

packages/eslint-config-angular/template.js
m-akinc: Now that all these rules have been collected into `@angular-eslint/template/accessibility`, I'm following the pattern we used for `@angular-eslint/template/recommended`, i.e. one identifying comment before this group of overrides. I could add the `[accessibility]` tag to that comment, or update the doc so it doesn't reference a tag.

I should note that Angular doesn't classify `no-positive-tabindex` as an accessibility rule, so I removed the tag there, too.

rajsite: Don't think I follow your response from the readme:

```
Evaluate project-specific rule groups
Several sets of rules may be enabled based on requirements of a given project. By default the following sets of rules are in an inert / disabled state, but should be evaluated for your integration.

Text search for the tag associated with a specific rule group in the repository to find the related rules. If enabling a rule group, the rules should be toggled from 'off' to 'error' unless the rule comment says otherwise.
```

And then the section of the README I linked:

```
Tag: [[accessibility]](https://github.com/ni/javascript-styleguide/search?q=%5Baccessibility%5D&type=code)

There currently isn't an NI organization wide requirement to enforce accessibility in applications. The rule group should be enabled if individual applications prioritize accessibility.
```

This PR deleted all the accessibility tags for some reason. So now there is nothing in that rule group for apps to choose to enable. Searching for that tag shows no results:

image

m-akinc: It's fine, I can put all the tags back. I'll also go back to not extending [`@angular-eslint/template/accessibility`](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/src/configs/README.md#accessibility), and instead we'll list each rule explicitly. I do think it makes sense to group these accessibility rules together, though, under a comment that links to the premade accessibility config.

rajsite: > I should note that angular-eslint doesn't classify no-positive-tabindex as an accessibility rule, so I removed the tag there, too.

The tags are about telling apps what accessibility rules exist and dealing with positive tab indexes is definitely accessibility related: https://dequeuniversity.com/tips/tabindex-positive-numbers

rajsite: re-opening comment since looks like there was a race on the last comment

m-akinc: It's weird (and kind of annoying) that they didn't include `no-positive-tabindex` in their accessibility preset. It might be a mistake:
image

I'll add the tag and put it with the other accessibility rules.

rajsite: > I do think it makes sense to group these accessibility rules together, though, under a comment that links to the premade accessibility config.

I'm not necessarily asking for that, just to make it clear how to make accessibility rules discoverable to apps based on the docs. I pointed out your change made it so the README made the rules not easy to discover. Reverting the change is an acceptable alternative. I offered to chat offline to try and understand what you think is clearer. But sounds like you would prefer not to engage in that discussion which is fine too.

change/@ni-eslint-config-angular-7bcf1406-f6fc-4336-a2ff-a2fbb32de903.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Just a note that we decided not to complete this PR until key clients are ready to uptake it.

jattasNI (approved):

change/@ni-eslint-config-angular-7bcf1406-f6fc-4336-a2ff-a2fbb32de903.json
TrevorKarjanis: There's really that much concern? Shouldn't this be much easier/faster than the Material upgrade?

jattasNI: We're just saying that it's more expedient to wait for client teams to complete their Angular 16 upgrade testing before we complete this PR. We could complete it earlier but didn't see any benefit to doing so, just churn of Renovate PRs that would have to be ignored / rejected until the Angular upgrade is ready.

TrevorKarjanis: Ohh, I misunderstood. Thanks for clarifying.

rajsite:
@m-akinc, don't need to wait for @mure since there should effectively be no rule change. Let me know when ready for the Angular 16 release train and I'll merge.

mure (approved):

PR #140, opened 2024-05-03
Update @angular-eslint/template/i18n to ignore common nimble and sl-lib attributes

### LINKED ISSUE #83: @angular-eslint/template/i18n flags attributes that don't need to be localized
In [AzDO PR 256885](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/256885) we saw an issue where [@angular-eslint/template/i18n](https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/docs/rules/i18n.md) flagged an attribute as needing localization when it wasn't actually necessary.

In that case, the template looked like this:
```html
<nimble-theme-provider theme="light">
```

The rule reported that we should add an additional `il8n-theme` attribute to mark the `theme`'s value to be localized. That seems wrong, since the name of the theme is not user-visible. The [Angular docs for localizing attributes](https://angular.io/guide/i18n-common-prepare#mark-element-attributes-for-translations) explain how this works, but don't explicitly say that every attribute must be translated.

Here are the configuration options available for the rule:
```ts
interface Options {
  boundTextAllowedPattern?: string;
  /**
   * Default: `true`
   */
  checkAttributes?: boolean;
  /**
   * Default: `true`
   */
  checkId?: boolean;
  /**
   * Default: `true`
   */
  checkText?: boolean;
  /**
   * Default: `["charset","class","color","colspan","fill","formControlName","height","href","id","lang","ngClass","ngProjectAs","routerLink","src","stroke","stroke-width","style","svgIcon","tabindex","target","type","viewBox","width","xmlns"]`
   */
  ignoreAttributes?: string[];
  ignoreTags?: string[];
  requireDescription?: boolean;
}
```

Some possible remedies:
1. Set `checkAttributes` to `false`
2. Add `theme` to `ignoreAttributes`
3. Add `nimble-theme-provider` to `ignoreTags`
4. Add docs providing guidance for when projects should set one of the above configurations, but keep current config the same
5. Add docs about when inline suppressions are appropriate (note that inline suppressions in HTML are kind of noisy: `<!-- eslint-disable-next-line @angular-eslint/template/i18n -->`)
6. Other thoughts?

rbell517:
## Justification

Update `@angular-eslint/template/i18n` to ignore common nimble and sl-lib attributes (and excludes tests of the name `*.spec.ts` with inline templates).

fixes #83

## Implementation

A new file `template/options.js` is added to list all well-known attributes that should not be localized for using in NI javascript codebases. This includes common nimble attributes, systemlink-lib-angular attributes, and attributes from commonly used third-party libraries.

This list is used to modify the `@angular-eslint/template/i18n` eslint rule to ignore any attribute in the list by default. If apps need to change this to ignore only nimble attributes, they have the option to do that by overriding the rule's config in their app's eslintrc.

## Testing

- I have made modifications to the angular test project to include attributes on the two sample templates that should either be ignored or not ignored and verified the eslint rule modification is being correctly applied in each.
- Added a test to verify custom attributes can be added to ignore lists by apps.

### DIFF
--- a/change/@ni-eslint-config-angular-ff70c9b2-14ea-4eb3-9c0e-0de3cdb388ec.json
+++ b/change/@ni-eslint-config-angular-ff70c9b2-14ea-4eb3-9c0e-0de3cdb388ec.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update @angular-eslint/template/i18n to ignore common nimble and sl-lib attributes",
+  "packageName": "@ni/eslint-config-angular",
+  "email": "418101+rbell517@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/eslint-config-angular/i18n-attributes-ignore.js
+++ b/packages/eslint-config-angular/i18n-attributes-ignore.js
@@ -1,4 +1,16 @@
-module.exports = {
+// This is a list of standard element attributes to ignore
+// when applying the @angular-eslint/template/i18n rule
+// that requires attributes to be tagged as localizable.
+//
+// New attributes can be added here that are for element
+// configuration and not for display, where localizing
+// the attribute value would result in broken functionality.
+//
+// In cases of attribute collisions, the rule supports
+// CSS selector style element[attribute] ignore entries
+// which can be used to scope attribute ignores to specific
+// elements.
+const ignoreAttributes = {
     nimble: [
         'action-menu-slot',
         'appearance',
@@ -15,7 +27,7 @@ module.exports = {
         'slot',
         'theme'
     ],
-    'systemlink-lib-angular': [
+    systemlink: [
         // sl-workspace-selector
         'action',
 
@@ -47,4 +59,9 @@ module.exports = {
         'matColumnDef',
         'matTooltipClass',
     ]
+};
+
+module.exports = {
+    ...ignoreAttributes,
+    all: [...ignoreAttributes.nimble, ...ignoreAttributes.systemlink, ...ignoreAttributes.jqx, ...ignoreAttributes.material]
 };
--- a/packages/eslint-config-angular/i18n-attributes-ignore.js
+++ b/packages/eslint-config-angular/i18n-attributes-ignore.js
@@ -1,0 +1,50 @@
+module.exports = {
+    nimble: [
+        'action-menu-slot',
+        'appearance',
+        'appearance-variant',
+        'column-id',
+        'field-name',
+        'format',
+        'icon',
+        'id-field-name',
+        'key',
+        'key-type',
+        'orientation',
+        'severity',
+        'slot',
+        'theme'
+    ],
+    'systemlink-lib-angular': [
+        // sl-workspace-selector
+        'action',
+
+        // sl-table
+        'columnId',
+        'fieldName',
+        'idFieldName',
+        'selectionMode',
+        'slTableColumnId',
+
+        // sl-grid
+        'dataRowId',
+        'loadColumnStateBehavior',
+        'parentDataField'
+    ],
+    jqx: [
+        // smart-table
+        'sortMode',
+
+        // smart-query-builder
+        'applyMode',
+        'fieldsMode'
+    ],
+    material: [
+        'cdkDragPreviewContainer',
+        'floatLabel',
+        'fontIcon',
+        'fontSet',
+        'matColumnDef',
+        'matTooltipClass',
+    ]
+};
--- a/packages/eslint-config-angular/options.js
+++ b/packages/eslint-config-angular/options.js
@@ -60,8 +60,8 @@ const ignoreAttributes = {
         'matTooltipClass',
     ]
 };
+ignoreAttributes.all = Object.values(ignoreAttributes).flat();
 
 module.exports = {
-    ...ignoreAttributes,
-    all: [...ignoreAttributes.nimble, ...ignoreAttributes.systemlink, ...ignoreAttributes.jqx, ...ignoreAttributes.material]
+    ignoreAttributes
 };
--- a/packages/eslint-config-angular/template.js
+++ b/packages/eslint-config-angular/template.js
@@ -74,7 +74,7 @@ module.exports = {
             'error',
             {
                 checkId: false,
-                ignoreAttributes: [...ignoreAttributes.nimble, ...ignoreAttributes['systemlink-lib-angular'], ...ignoreAttributes.jqx, ...ignoreAttributes.material]
+                ignoreAttributes: [ignoreAttributes.all]
             }
         ],
 
--- a/packages/eslint-config-angular/template.js
+++ b/packages/eslint-config-angular/template.js
@@ -1,4 +1,4 @@
-const ignoreAttributes = require('./i18n-attributes-ignore');
+const { ignoreAttributes } = require('./options');
 
 module.exports = {
     extends: [
@@ -94,5 +94,18 @@ module.exports = {
             Providing a `trackBy` function in `ngFor` loops can improve performance in specific cases where Angular can't track references, but it's overkill to require it for every `ngFor` so this rule is disabled.
         */
         '@angular-eslint/template/use-track-by-function': 'off'
-    }
+    },
+    overrides: [
+        {
+            // Ignore inline templates in tests using the inline template naming convention
+            // See naming details: https://github.com/angular-eslint/angular-eslint/releases/tag/v14.0.0
+            files: ['*.spec.ts*.html'],
+            rules: {
+                /*
+                    Tests often define helper components that don't need to be marked for i18n.
+                */
+                '@angular-eslint/template/i18n': 'off'
+            }
+        },
+    ]
 };
--- a/packages/eslint-config-angular/template.js
+++ b/packages/eslint-config-angular/template.js
@@ -1,3 +1,5 @@
+const ignoreAttributes = require('./i18n-attributes-ignore');
+
 module.exports = {
     extends: [
         'plugin:@angular-eslint/template/recommended'
@@ -68,7 +70,13 @@ module.exports = {
             and new applications in the chance that they'll need to be localized in the future. Disable this rule
             if an application will never be localized.
         */
-        '@angular-eslint/template/i18n': ['error', { checkId: false }],
+        '@angular-eslint/template/i18n': [
+            'error',
+            {
+                checkId: false,
+                ignoreAttributes: [...ignoreAttributes.nimble, ...ignoreAttributes['systemlink-lib-angular'], ...ignoreAttributes.jqx, ...ignoreAttributes.material]
+            }
+        ],
 
         '@angular-eslint/template/no-any': 'error',
 
--- a/packages/eslint-config-angular/template.js
+++ b/packages/eslint-config-angular/template.js
@@ -1,4 +1,4 @@
-const { ignoreAttributes } = require('./options');
+const { ignoreAttributes } = require('./template/options');
 
 module.exports = {
     extends: [
--- a/packages/eslint-config-angular/template.js
+++ b/packages/eslint-config-angular/template.js
@@ -74,7 +74,7 @@ module.exports = {
             'error',
             {
                 checkId: false,
-                ignoreAttributes: [ignoreAttributes.all]
+                ignoreAttributes: [...ignoreAttributes.all]
             }
         ],
 
--- a/tests/angular/custom-ignore-attributes/.eslintrc.js
+++ b/tests/angular/custom-ignore-attributes/.eslintrc.js
@@ -1,4 +1,4 @@
-const {ignoreAttributes} = require('@ni/eslint-config-angular/options');
+const {ignoreAttributes} = require('@ni/eslint-config-angular/template/options');
 
 module.exports = {
     overrides: [{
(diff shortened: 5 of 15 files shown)


change/@ni-eslint-config-angular-ff70c9b2-14ea-4eb3-9c0e-0de3cdb388ec.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update @angular-eslint/template/i18n to ignore common nimble and sl-lib attributes",
+  "packageName": "@ni/eslint-config-angular",
+  "email": "richard.bell@ni.com",
```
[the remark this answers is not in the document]
rbell517: Thanks for the tip

packages/eslint-config-angular/i18n-attributes-ignore.js
```diff
@@ -0,0 +1,50 @@
...
+        'fieldsMode'
+    ],
+    material: [
+        'cdkDragPreviewContainer',
+        'floatLabel',
+        'fontIcon',
+        'fontSet',
+        'matColumnDef',
+        'matTooltipClass',
+    ]
```
rajsite, line 49: Are we loosing the DEFAULT_ALLOWED_ATTRIBUTES ?

https://github.com/angular-eslint/angular-eslint/blob/20bbfb20ef0a0e4d76b7482dc8263492c819e9f2/packages/eslint-plugin-template/src/rules/i18n.ts#L112C3-L112C19

If so how should we manage them, doesn't look like they are exported.. copy and paste in with link back?

packages/eslint-config-angular/i18n-attributes-ignore.js
```diff
@@ -0,0 +1,50 @@
...
+        'icon',
+        'id-field-name',
+        'key',
+        'key-type',
+        'orientation',
+        'severity',
+        'slot',
+        'theme'
+    ],
+    'systemlink-lib-angular': [
```
rajsite, line 18: What about just `systemlink` since these are really domains more than exact packages.

```suggestion
    systemlink: [
```

I'd also be up for exact packages. `@ni/nimble-angular`, `@ni-kismet/systemlink-lib-angular`, etc

packages/eslint-config-angular/i18n-attributes-ignore.js
rbell517: No, we aren't losing them. Lower in that file you can see they [concat the two together](https://github.com/angular-eslint/angular-eslint/blob/20bbfb20ef0a0e4d76b7482dc8263492c819e9f2/packages/eslint-plugin-template/src/rules/i18n.ts#L219). And you can see in the test I added the i18n tag is still needed for a `title` attribute.

packages/eslint-config-angular/template.js
```diff
@@ -68,7 +70,13 @@ module.exports = {
             and new applications in the chance that they'll need to be localized in the future. Disable this rule
             if an application will never be localized.
         */
-        '@angular-eslint/template/i18n': ['error', { checkId: false }],
+        '@angular-eslint/template/i18n': [
+            'error',
+            {
+                checkId: false,
+                ignoreAttributes: [...ignoreAttributes.nimble, ...ignoreAttributes['systemlink-lib-angular'], ...ignoreAttributes.jqx, ...ignoreAttributes.material]
```
rajsite, line 77: If apps choose to augment the list and we add more categories seems like it would be easy to get out of sync.

Speculating but I think it's unlikely folks will be really concerned about removing things from the list. But having the full list easily available to easily append to could be useful. Should we have that as an export and use it here?

packages/eslint-config-angular/i18n-attributes-ignore.js
rajsite: Scanning that source it does look like we could scope these to specific elements: https://github.com/angular-eslint/angular-eslint/blob/20bbfb20ef0a0e4d76b7482dc8263492c819e9f2/packages/eslint-plugin-template/src/rules/i18n.ts#L525

Not sure that is super valuable but it is an option. Might be useful to have a comment on the top of the list describing the recommendation and stating element specific attributes are possible

rbell517: I've update the attribute list to include a block comment that describes the list's purpose, the intent that should be followed when adding to it, and mentions to option to use `element[attribute]` notation for more specific scoping.

packages/eslint-config-angular/i18n-attributes-ignore.js
rbell517: Change to just `systemlink`

packages/eslint-config-angular/template.js
rbell517: I've moved the attribute lists into a const to allow adding an `all` option to the module exports that concats all 4 together, and used that `all` option here.

jattasNI (approved):
You can complete this once @rajsite approves, no need to wait on the remaining reviewers.

tests/angular/custom-ignore-attributes/.eslintrc.js
```diff
@@ -0,0 +1,21 @@
+const {ignoreAttributes} = require('@ni/eslint-config-angular/options');
```
rajsite, line 1: This is a new public api so added a test for it. Updated the export to be from a generic `/options` entry point

Was on the fence about breaking up the options more, like should we

[1] export config for:
`@ni/eslint-config-angular` from `@ni/eslint-config-angular/options`
and
`@ni/eslint-config-angular/template` from `@ni/eslint-config-angular/template/options`?

[2] Or just export every option from: `@ni/eslint-config-angular/options` as this PR now currently does.

I think I kinda like [1] more which is an idea I had after pushing the change but wanted feedback first. @jattasNI

tests/angular/custom-ignore-attributes/.eslintrc.js
```diff
@@ -0,0 +1,21 @@
...
+                    ignoreAttributes: [...ignoreAttributes.all, 'custom-field']
+                }
+            ],
+        }
+    }, {
+        files: ['*.spec.ts*.html'],
+        rules: {
+            '@angular-eslint/template/i18n': 'off'
+        }
+    }]
```
rajsite, line 20: If an app wants to add additional custom attributes to ignore i18n this is what it looks like. It's kinda annoying to have to re-disable the checks in tests but maybe that's fine? :/

@jattasNI @rbell517  thoughts on if we should document this pattern for ignoring project-specific attributes? I'm not really sure how often it'll come-up. Should we pre-emptively run local builds of the lints in more projects to see if many apps will need overrides? If they do should we try and come up with better ways to configure this?

packages/eslint-config-angular/options.js
```diff
@@ -0,0 +1,67 @@
...
+//
+// New attributes can be added here that are for element
+// configuration and not for display, where localizing
+// the attribute value would result in broken functionality.
+//
+// In cases of attribute collisions, the rule supports
+// CSS selector style element[attribute] ignore entries
+// which can be used to scope attribute ignores to specific
+// elements.
+const ignoreAttributes = {
```
rajsite, line 13: @jattasNI we have this rule disable in nimble: https://github.com/ni/nimble/blob/4e595cc0e67a9903510ba60da17afdb1bf2643d6/packages/angular-workspace/.eslintrc.js#L109

Maybe we should pre-emptively enable it to see if the example app reveals anymore attributes that need to be marked? How confident are we in the completeness of the current list / should we do any work preemptively to shore-it up for nimble?

packages/eslint-config-angular/template.js
```diff
@@ -86,5 +94,18 @@ module.exports = {
             Providing a `trackBy` function in `ngFor` loops can improve performance in specific cases where Angular can't track references, but it's overkill to require it for every `ngFor` so this rule is disabled.
         */
         '@angular-eslint/template/use-track-by-function': 'off'
-    }
+    },
+    overrides: [
```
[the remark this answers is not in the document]
rbell517: Yes I think a post mentioning this change, mentioning what changes to apps are recommended, and reminding authors and reviewers to be observant for existing i18n tags on attributes that should not be localized, because this change does nothing to flag existing or new instances of that. 

I'll plan to make that post.

tests/angular/custom-ignore-attributes/.eslintrc.js
rbell517: I personally don't think we need to document anything for this. The tests themselves are reasonable documentation and realistically its most likely that if there are extra attributes to ignore they are going to come from a library and we would want to add to our ignore list instead.

tests/angular/custom-ignore-attributes/.eslintrc.js
jattasNI: I kinda like the symmetry of 1. If needed in the future we could also add '@ni/eslint-config-angular/requiring-type-checking/options'.

tests/angular/custom-ignore-attributes/.eslintrc.js
jattasNI: I'm ok to wait until someone asks how to do it.

packages/eslint-config-angular/options.js
jattasNI: I am not very confident in the completeness of the list. In my [innovation day prototype PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/667662?_a=files&path=/Web/Workspaces/SystemLinkShared/projects/systemlink-lib-angular/i18n-tags.js) I generated it by trial and error on the SystemLinkShared libraries so it could easily be missing APIs that are used by apps (or by nobody yet) if they're not used by a shared library. I think this PR largely uses that list, so it's probably incomplete.

I am fine with making the list comprehensive either before or after completing this PR. I suspect that adopting the package in SLE apps will reveal more APIs than adopting it in the Nimble example app since the latter doesn't exercise APIs very comprehensively. But creating a local package and locally adopting it in every app is a lot of busy work so if @rbell517 would rather wait for Renovate to do that work on a production build and then add any missing attributes as a follow up PR, that's ok by me.

Similar thought on enabling i18n on the example app. I don't object to giving it a shot but I think it'll miss a lot of new APIs.

jattasNI (approved):

packages/eslint-config-angular/options.js
rajsite: > But creating a local package and locally adopting it in every app is a lot of busy work so if @rbell517 would rather wait for Renovate to do that work on a production build and then add any missing attributes as a follow up PR, that's ok by me.

If @rbell517 is okay keeping an eye out on rennovate PRs and helping message / coral folks running into issues post merge that works for me 👍

> I suspect that adopting the package in SLE apps will reveal more APIs than adopting it in the Nimble example app since the latter doesn't exercise APIs very comprehensively.

Yea that's true. The test suite is intended to test the angular bindings exhaustively though. Maybe we should turn on i18n in nimble-angular's test suite 🤔

tests/angular/custom-ignore-attributes/.eslintrc.js
rajsite: Yea true. Lets go with [1]. I'll push a change for it.

packages/eslint-config-angular/template.js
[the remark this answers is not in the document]
rajsite: sounds good to me, thanks @rbell517 👍

TrevorKarjanis (approved):

packages/eslint-config-angular/options.js
rajsite: @rbell517 if you are okay with this direction then we are ready to ship

rajsite: Offline sounds like @rbell517 has additional feedback to give so cancelling autocomplete to prevent accidental release.

rbell517: Just so there are not any misunderstandings on the impact of this change, uptaking this new version in libraries and apps will not by itself "reveal more APIs." This change means errors will no longer be generated for the listed attributes when it doesn't have an i18n attribute, but it will not generate errors when the listed attributes do have an i18n attribute. That is just a limitation of the eslint rule. It also does nothing to the standard `[attribute]="'value'"` trick to avoid the eslint rule without suppressing. The net is this is strictly a loosing of the rule, no new errors should be generated by uptaking the new version, so there shouldn't be any need to monitor Renovate builds.

What I've done instead is use some regex look for cases we worked around the i18n rule (`\[[a-zA-Z-]+\]="'.*'"`) or have added a bad i18n- attribute (`i18n-[a-zA-Z-]+`) and scrubbed the results for SystemLinkShared, TestInsights, LabManagement, and SystemsManagement. This revealed 5 instances of bad i18n attributes (all Lab Management), which is actually much better than I feared. It revealed 15 more attributes to add to the ignore list, which I've pushed to this PR. Most of these are `sl-table` and `sl-grid` related, but there are also 6 new nimble attributes. Between those 4 workspaces I think we will have close to complete coverage. Please give these a look over @jattasNI @rajsite and resolve this if you're satisfied.

rbell517: Correction, only 3 instances of bad i18n- attributes. I mis-identified one and the other was already fixed.
https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/715355

rajsite: It's strictly a loosening makes sense, that hadn't computed for me. Also justifies the patch revision 👍

Sounds like part of the discussion for the UI Working Group should be to search for and remove those workarounds and update the list. Might even be tech debt tasks to farm out but that could be a working group discussion. LGTM!

rajsite (approved):

jattasNI:
This [made it into SystemLinkShared](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/715410) with no pipeline failures 🎉

PR #146, opened 2024-05-18
More nimble i18n attribute exceptions

rajsite:
More i18n attribute exceptions found by enabling the attributes part of the i18n rule in nimble, see: https://github.com/ni/nimble/pull/2122

### DIFF
--- a/change/@ni-eslint-config-angular-8697df94-b254-47fd-813b-d9f9722037b4.json
+++ b/change/@ni-eslint-config-angular-8697df94-b254-47fd-813b-d9f9722037b4.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add more nimble i18n attr exceptions",
+  "packageName": "@ni/eslint-config-angular",
+  "email": "rajsite@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/eslint-config-angular/template/options.js
+++ b/packages/eslint-config-angular/template/options.js
@@ -12,40 +12,40 @@
 // elements.
 const ignoreAttributeSets = {
     web: [
+        'aria-describedby',
         'aria-labelledby',
         'rel',
         'sizes',
-        'slot',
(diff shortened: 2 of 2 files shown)


change/@ni-eslint-config-angular-8697df94-b254-47fd-813b-d9f9722037b4.json
```diff
@@ -0,0 +1,7 @@
+{
```
rajsite, line 1: working on a couple more tweaks

packages/eslint-config-angular/template/options.js
```diff
@@ -11,29 +11,54 @@
...
         'format',
+        'location',
+        'orientation',
+        'resize',
+        'severity',
+        'theme',
+
+        // rich text
+        'button-label',
+        'display-name',
```
jattasNI, line 40: Looks like both of these end up as user visible strings. Generally we would want to remove them from this list so that the rule encourages them to be localized.

I think that'd be the right approach for `button-label`: it's the tooltip for the `@` button in the rich text toolbar so if someone provides it, they should provide a localized string.

Looks like `display-name` is the visible name of a mentioned user. I suspect we don't want those to be localized so keeping this attribute in the list seems correct.

TrevorKarjanis (approved):

packages/eslint-config-angular/template/options.js
```diff
@@ -11,29 +11,54 @@
 // which can be used to scope attribute ignores to specific
 // elements.
 const ignoreAttributeSets = {
+    web: [
```
jattasNI, line 14: Can you move `accept` and `aria-live` from down below into this new section?

jattasNI (approved):

packages/eslint-config-angular/template/options.js
rajsite: Removed button-label and because display-name feels a little ambiguous I scoped it to `nimble-mapping-user`

PR #149, opened 2024-07-12
Menu button column i18n exceptions

mollykreis:
The `nimble-table-column-menu-button` has a `menu-slot` attribute that should not be localized. Therefore, it needs to be added to the list of attribute exceptions.


rajsite:
@mollykreis released!

TrevorKarjanis (approved):

PR #152, opened 2024-08-08
Exempt parentIdFieldName from i18n rule

m-akinc:
The `parentIdFieldName` attribute of the `sl-table` does not require translation, so add it to the exemptions.


rajsite (approved):

rajsite:
FYI @m-akinc released 🎉

Issue #73, opened 2021-12-01
Make it easier to adopt strictNullCheck rules


rajsite:
Right now [the topic](https://github.com/ni/javascript-styleguide/blob/059ceded256abc716ce682abdd03b8c1d8aee159/README.md#strict-null-checks) links to a search for the [strict-null-checks] flag but when you visit those spots in source it is not immediately clear what the recommended config with strict null checks enabled should be (just guessed that 'off' should toggle to 'error', etc, but that might not be clearer to first time eslint users).

We can either put recommended configs in the comments or go one step further and create strict versions of the rules, ie:
- @ni/eslint-config-typescript/strict
- @ni/eslint-config-typescript/requiring-type-checking-strict
- @ni/eslint-config-angular/strict
- @ni/eslint-config-angular/requiring-type-checking-strict

I think having dedicated rulesets for strict mode will be handy to encourage consistent strict-mode usage in apps and make it easier to adjust the strict mode configuration going forward. Will also make it easier for new projects trying to start with good strict mode settings.

rajsite:
Alternate:

Default as strict and make current configs as:

- @ni/eslint-config-typescript/less-strict
- @ni/eslint-config-typescript/requiring-type-checking-less-strict
- @ni/eslint-config-angular/less-strict
- @ni/eslint-config-angular/requiring-type-checking-less-strict

rajsite:
With strictNullChecks the eslint default-case and typescript-eslint/switch-exhaustiveness-check [are incompatible](https://github.com/typescript-eslint/typescript-eslint/issues/2959). Should the strict ruleset disable default-case in favor of switch-exhaustiveness-check?

rajsite:
Re-upping this issue for 2023 now that we've have a bit of runtime and use strict mode in more projects.

New proposal:

1. Make our default configuration files strict (encourage strict rules as the default)
2. Merge the requiring-type-checking rules into the default rules (doing code searches it does not seem like repos opt out of the requiring-type-checking rule sets)
3. Add the following configuration files which extend the default files:
  - @ni/eslint-config-typescript/legacy
  - @ni/eslint-config-angular/legacy

So the end result would be the following configs:
  - @ni/eslint-config-typescript
  - @ni/eslint-config-typescript/legacy
  - @ni/eslint-config-angular
  - @ni/eslint-config-angular/legacy

The intentions are to make the out-of-the-box configurations extending the base rules strict and aligned by default.
[...]

PR #156, opened 2024-09-19
Async rule changes

### LINKED ISSUE #115: Consider enabling the require-await rule
## Overview

I know, I know.. hear me out. I found that the [require-await](https://eslint.org/docs/latest/rules/require-await) rule was disabled in nimble and think it actually might make sense to enable these days. I can't say I've fully investigated it but I found it useful enough to [enable in nimble](https://github.com/ni/nimble/pull/833).

### Reduce error code paths
Our current config comes from airbnb where the [commit comments](https://github.com/airbnb/javascript/commit/0b3b5a178fb32bfc60096738369e069c8cfcfa91#r35614411) give more insight than the description in the source file.

I'd summarize the airbnb logic as they like async-await because if a dev wrote something like:

```js
function myFunc () {
    if (woops) {throw new Error}
    return new Promise((resolve, reject) => reject());
}
```

They have a function that has multiple error code paths: A synchronous exception and an asynchronous rejection to handle. However if you make the function async then you only have to deal with the asynchronous rejection error path. The throw statement would always turn into an async rejection:

```js
async function myFunc () {
    if (woops) {throw new Error}
    return new Promise((resolve, reject) => reject());
}
```

For a funtion that returns a promise, making that function async helps reduce the error cases to consider.

### Improve performance

When the aribnb rule was made, the behavior of awaiting a promise in an async function had signficant overhead:

The following (where bar returns a promise): 

```js
async function foo() {
  const result = await bar(); // result is not a promise, it is the resolved value
  return result; // result needs to be wrapped in a promise
}
```

Was less performant than (where bar still returns a promise):

```js
async function foo() {
  const result = bar(); // result is a non-resolved promise
  return result; // the promise can pass through without a wrapper
}
```

### Avoid unnecessary async

The immediate use case I found was that functions in the nimble code base were being made [unnecessarily async](https://github.com/ni/nimble/pull/831#discussion_r1029679411). While that example is innocuous, making functions unnecessarily async is only a detriment to both performance and code complexity. Having the rule enabled lets the linter flag unnecessarily async cases.

### Consistent async patterns keep error handling consistent

At least in nimble the reviewer recommendation (not official / documented) is to avoid mixing Promise and async/await style code. What this means in practice is that the airbnb example above would be refactored to something like the following in PR:

```js
function doSomething() {
    return new Promise((resolve, reject) => reject());
}

async function myFunc () {
    if (woops) {throw new Error}
    await doSomething();
}
```

The doSomething function only uses Promises and the myFunc function is only using async / await. In practice, this results in the promise functions doing minimal work outside the Promise constructor so synchrous exception code paths are unlikely and the async functions have much clearer behavior by only handling one form of async control flow at a time.

### Incidental performance concerns have been mitigated

The example given, i.e.:

```js
async function foo() {
  const result = await bar();
  return result;
}
```

Now has minimal overhead in modern engines (at least in V8). As discussed in the [Faster async functions and promises](https://v8.dev/blog/fast-async#fn2) actual spec changes were made that avoid the additional microtask from occuring with nested async-await statements and temporary promises are elided.

Compared to trying to manually write promise code to control microtask points and promise instancing, the article states: "async/await outperforms hand-written promise code now".

### Function resolution is at the call-site

With the example:

```js
async function foo() {
  const result = bar();
  return result;
}
```

The actual resolution of the invocation of the bar function is not in the value of result. It's conceptually and debugging wise pretty annoying. Setting a breakpoint at this location in code will only give you an unresolved promise.

However with the following:

```js
async function foo() {
  const result = await bar();
  return result;
}
```

The value of result is the resolved value in the function. This makes more sense in terms of control flow and helps during debugging. With the understanding that there is not a performance cost to it anymore, then it should be preferable to await closer to the call site that creates the promise.

## Summary

Assuming the style guideline that async/await and Promise styles of coding should not be mixed, then it probably makes sense today to enable the `require-await` rule.

### LINKED ISSUE #151: Enable no-floating-promises for Angular tests
**Describe the issue**

The rule [@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises/) is currently enabled for most files but [disabled for Angular tests](https://github.com/ni/javascript-styleguide/blob/6d3cf459ec54310f6e0a01254e8d2d540e00d789/packages/eslint-config-angular/requiring-type-checking.js#L14). We should consider removing that configuration so it is enabled for all code.

Rationale:
1. I recently found many instances of Angular tests that aren't handling promises returned from methods they call. (Here's [a draft PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/761645?_a=files&path=/Web/Workspaces/SystemLinkShared/projects/systemlink-lib-angular/notebook/src/components/sl-notebook-execution-form/sl-notebook-execution-form.component.spec.ts) that enables this and some other async rules in systemlink-lib-angular). I can't point to a specific instance of problems this causes but it seems likely to cause intermittency.
2. Our rationale for disabling it was that "The jasminewd2 library used by Angular applications results in a significant number of floating promises and unbound methods". I didn't see any instances of false positives like this in the PR above.

I recommend we try enabling it in a few places and evaluate the changes it requires. If we're happy with its suggestions we should enable it by default so that new projects get its benefits. If needed, apps can continue to disable it while they fix the issues it catches.

jattasNI:
# Rationale

Resolves #151. This makes it so a promise must be awaited (or otherwise handled, just not ignored) in Angular tests, making them match other code.

Resolves #115. This makes it so a function marked `async` must `await` something, not contain only synchronous code and not return a promise.

# Implementation

1. Stop disabling `no-floating-promises` in Angular tests. See PR comment for rationale for keeping `no-unbound-method: 'off'` here.
    - this repo even had some Angular tests that violated this rule because they were using stale `jasminewd2` types. I fixed those by using `jasmine` types instead (and fixing a playwright test that was incorrectly referencing jasmine types)
    - this required adding an `.npmrc` since my dev machine is configured to use Artifactory
3. Set `no-return-await: 'off'` in the JS ruleset. This [rule is now deprecated by ESLint](https://eslint.org/docs/latest/rules/no-return-await) because returning an awaited promise is no longer a performance concern.
    - the related `@typescript-eslint/return-await': 'error'` was already enabled and is still useful because [it enforces the inverse](https://typescript-eslint.io/rules/return-await/), that you should await promises rather than just return them because you get better stack traces
4. Set `'require-await': 'error'` in the JS ruleset and `'@typescript-eslint/require-await': 'error'` in the TS-with-typechecking ruleset.
5. Change files for beachball marked `minor` in accordance with CONTRIBUTING.md guidance for rule changes and include package-specific descriptions of the changes

# Testing

[A draft PR](https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/761645) into `systemlink-lib-angular` shows the impact of some of these changes. It requires mostly mechanical changes and catches lots of legitimate issues.

I also updated a couple test files in this repo to include more async cases and played around with those implementations locally to verify the rules were working.

### DIFF
--- a/packages/eslint-config-angular/requiring-type-checking.js
+++ b/packages/eslint-config-angular/requiring-type-checking.js
@@ -7,6 +7,10 @@ module.exports = {
         {
             files: ['*.spec.ts'],
             rules: {
+                /*
+                    Spies used by Angular application tests result in a significant number of
+                    unbound methods so this rule is disabled for test specs in Angular projects.
+                */
                 '@typescript-eslint/unbound-method': 'off',
             }
         },
--- a/packages/eslint-config-angular/requiring-type-checking.js
+++ b/packages/eslint-config-angular/requiring-type-checking.js
@@ -7,11 +7,6 @@ module.exports = {
         {
             files: ['*.spec.ts'],
             rules: {
-                /*
-                    The jasminewd2 library used by Angular applications results in a significant number of
-                    floating promises and unbound methods so these rules are disabled for test specs in Angular projects.
-                */
-                '@typescript-eslint/no-floating-promises': 'off',
                 '@typescript-eslint/unbound-method': 'off',
             }
         },
--- a/packages/eslint-config-javascript/index.js
+++ b/packages/eslint-config-javascript/index.js
@@ -325,6 +325,11 @@ module.exports = {
                 selector: 'CallExpression[callee.object.name=\'console\'][callee.property.name=/^(debug|info|time|timeEnd|trace)$/]',
                 message: 'Unexpected property on console object was called'
             }],
+        /*
+            This rule is deprecated since ESLint 8.46.0 because returning an awaited value no longer generates an extra microtask.
+            https://eslint.org/docs/latest/rules/no-return-await
+        */
+        'no-return-await': 'off',
 
         /*
             Disallow returning values from setters.
--- a/packages/eslint-config-javascript/index.js
+++ b/packages/eslint-config-javascript/index.js
@@ -429,6 +429,9 @@ module.exports = {
         */
         'prefer-regex-literals': 'error',
 
+        /*
+            Asynchronous functions that don’t use await might not need to be asynchronous functions and could be the unintentional result of refactoring.
+        */
         'require-await': 'error',
 
         /*
--- a/packages/eslint-config-javascript/index.js
+++ b/packages/eslint-config-javascript/index.js
@@ -424,6 +424,8 @@ module.exports = {
         */
         'prefer-regex-literals': 'error',
 
+        'require-await': 'error',
+
         /*
             This configuration already supports the JSDoc syntax. Add additional syntax as line or
             block exceptions or markers when necessary.
--- a/tests/angular/custom-ignore-attributes/index.spec.ts
+++ b/tests/angular/custom-ignore-attributes/index.spec.ts
@@ -18,7 +18,7 @@ describe('MyComponent', () => {
     it('should have a div', async () => {
         await fixture.whenStable();
 
-        expect(hostComponent.div).toBeDefined();
-        expect(hostComponent.myMethod).not.toHaveBeenCalled();
+        await expect(hostComponent.div).toBeDefined();
+        await expect(hostComponent.myMethod).not.toHaveBeenCalled();
     });
 });
--- a/tests/angular/custom-ignore-attributes/index.spec.ts
+++ b/tests/angular/custom-ignore-attributes/index.spec.ts
@@ -18,7 +18,7 @@ describe('MyComponent', () => {
     it('should have a div', async () => {
         await fixture.whenStable();
 
-        await expect(hostComponent.div).toBeDefined();
-        await expect(hostComponent.myMethod).not.toHaveBeenCalled();
+        expect(hostComponent.div).toBeDefined();
+        expect(hostComponent.myMethod).not.toHaveBeenCalled();
     });
 });
--- a/tests/javascript/index.js
+++ b/tests/javascript/index.js
@@ -12,4 +12,8 @@ export class NI {
     makeAwesomer() {
         this._awesomeLevel += 1;
     }
+
+    async asyncAwesomeness() {
+        await new Promise();
+    }
 }
--- a/tests/javascript/index.js
+++ b/tests/javascript/index.js
@@ -14,6 +14,6 @@ export class NI {
     }
 
     async asyncAwesomeness() {
-        await new Promise();
+        await Promise.resolve(42);
     }
 }
--- a/tests/typescript-requiring-type-checking/index.ts
+++ b/tests/typescript-requiring-type-checking/index.ts
@@ -12,8 +12,6 @@ export class NI {
     }
 
     public async asyncAwesomeness(): Promise<void> {
-        await new Promise(resolve => {
-            resolve(true);
-        });
+        await Promise.resolve(true);
     }
 }
--- a/tests/typescript-requiring-type-checking/index.ts
+++ b/tests/typescript-requiring-type-checking/index.ts
@@ -10,4 +10,10 @@ export class NI {
     public makeAwesomer(): void {
         this._awesomeLevel += 1;
     }
+
+    public async asyncAwesomeness(): Promise<void> {
+        await new Promise(resolve => {
+            resolve(true);
+        });
+    }
 }
--- a/tests/typescript-requiring-type-checking/index.ts
+++ b/tests/typescript-requiring-type-checking/index.ts
@@ -11,7 +11,9 @@ export class NI {
         this._awesomeLevel += 1;
     }
 
-    public async asyncAwesomeness(): Promise<void> {
-        await Promise.resolve(true);
+    public async slowAdd(a: number, b: number, timeMs: number): Promise<number> {
+        return await new Promise(resolve => {
+            setTimeout(() => resolve(a + b), timeMs);
+        });
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,5 @@ node_modules
 .DS_Store
 *.tgz
 .vscode
-.npmrc
 ./rules.txt
 ./rules-diff.txt
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+registry=https://registry.npmjs.org/
--- a/change/@ni-eslint-config-angular-225ce3da-b49d-4df7-ace0-b99abb8fbcb0.json
+++ b/change/@ni-eslint-config-angular-225ce3da-b49d-4df7-ace0-b99abb8fbcb0.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Rule changes for asynchronous code:\\n 1. Functions marked `async` must now include an `await`. The rule `'@typescript-eslint/require-await'` changed from `'off'` to `'error'`.\\n 2. Test code must now correctly handle promises. The rule `'@typescript-eslint/no-floating-promises'` changed from `'off'` to `'error'`.",
+  "packageName": "@ni/eslint-config-angular",
+  "email": "jattasNI@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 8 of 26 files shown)


tests/angular/custom-ignore-attributes/index.spec.ts
```diff
@@ -18,7 +18,7 @@ describe('MyComponent', () => {
     it('should have a div', async () => {
         await fixture.whenStable();
 
-        expect(hostComponent.div).toBeDefined();
-        expect(hostComponent.myMethod).not.toHaveBeenCalled();
+        await expect(hostComponent.div).toBeDefined();
+        await expect(hostComponent.myMethod).not.toHaveBeenCalled();
```
rajsite, line 22: expect isn't async right? There is a separate [expectAsync](https://jasmine.github.io/api/5.3/global.html#expectAsync)

jattasNI: Good catch, I had just mindlessly added the await. There was a lot going on here:

1. this project was importing `@types/jasminewd2` to get the jasmine types, and its matchers (e.g. `toBeDefined()` were async (the whole reason we had disabled this rule for tests in the first place)
2. jasminewd2 is from the Protractor days and no longer a dependency of Angular but this repo and Skyline still (incorrectly) reference it in a few spots

The fix was to remove it and replace it with `@types/jasmine`, which has synchronous matchers. This allows the tests to pass linting without the `await`.

While doing this I noticed that the playwright tests were somehow relying on the jasmine definition of `describe` so I updated one of them to use `test.describe` which is how Playwright does this.

I will make Skyline PRs to remove jasminewd2 as well.

jattasNI: https://dev.azure.com/ni/DevCentral/_git/Skyline/pullrequest/801169

packages/eslint-config-javascript/index.js
```diff
@@ -424,6 +429,8 @@ module.exports = {
         */
         'prefer-regex-literals': 'error',
 
+        'require-await': 'error',
```
rajsite, line 432: We are pretty consistent in this file with giving a reason in comment. Maybe (from the eslint docs directly): Asynchronous functions that don’t use await might not need to be asynchronous functions and could be the unintentional result of refactoring.

tests/javascript/index.js
```diff
@@ -12,4 +12,8 @@ export class NI {
     makeAwesomer() {
         this._awesomeLevel += 1;
     }
+
+    async asyncAwesomeness() {
+        await new Promise();
```
rajsite, line 17: nit / doesnt really matter, but seeing a promise that will never resolve just feels weird
```suggestion
        await Promise.resolve(42);
```

tests/typescript-requiring-type-checking/index.ts
```diff
@@ -10,4 +10,10 @@ export class NI {
     public makeAwesomer(): void {
         this._awesomeLevel += 1;
     }
+
+    public async asyncAwesomeness(): Promise<void> {
+        await new Promise(resolve => {
+            resolve(true);
+        });
```
rajsite, line 17:
```suggestion
        await Promise.resolve(true);
```

tests/typescript-requiring-type-checking/index.ts
```diff
@@ -10,4 +10,10 @@ export class NI {
     public makeAwesomer(): void {
         this._awesomeLevel += 1;
     }
+
+    public async asyncAwesomeness(): Promise<void> {
+        await new Promise(resolve => {
+            resolve(true);
```
rajsite, line 16: It's interesting that we don't end up catching this, but may be an artifact of direct Promise usage being harder to type check. But the types don't match, the function signature says `Promise<void>` but it's actually `Promise<boolean>`.

In practice if this was needed for some reason we should just do:

```ts
public async asyncAwesomeness(): Promise<boolean> {
    return await true;
}
```

To minimize ever using `Promise` directly and prefer `async-await`

Or alternatively (but less desirable):

```ts
public async asyncAwesomeness(): Promise<boolean> {
    return Promise.resolve(true);
}
```

Hopefully both of which are better typed.

rajsite: Actually I'm surprised either way that this isn't a type error. Maybe strict type checking isn't enabled here?

rajsite (approved):

tests/typescript-requiring-type-checking/index.ts
jattasNI: This comment revealed many interesting bits:
1. we don't set `strict: true` right now and in fact we set `strictNullChecks: false` because that's the default configuration for our ruleset, at least until we fix #73. 

2. That said, I don't think `strict` has anything to do with type errors like this being caught. Locally I do see a red squiggly if I create a type mismatch, but `npm run test` and `npm run lint` both succeed. Nothing in the pipeline ever does a TypeScript compile right now, so type errors are allowed in our tests I guess! 😅

3. The return type of this snippet (and the original more convoluted snippet) is correctly `Promise<void>` since it doesn't return the result of the await. It's a bit subtle, but it's allowed by our rules. Do you see any problem with that? I'm not sure of a way to block it.
```ts
public async asyncAwesomeness(): Promise<void> {
    await Promise.resolve(true);
}
```

4. I hadn't fully understood the [configuration options for @typescript-eslint/return-await](https://typescript-eslint.io/rules/return-await/#options) until now. Do we have a preference between these? According to the table in the rule docs, when outside a try-catch they are functionally equivalent and it's just a style difference. The current decision recorded in #115 is to stick with the default config, `in-try-catch`

```ts
    // allowed by 'always', blocked by 'in-try-catch'
    public async asyncAwesomenessA(): Promise<boolean> {
        return await Promise.resolve(true);
    }

    // blocked by 'always', allowed by 'in-try-catch'
    public async asyncAwesomenessB(): Promise<boolean> {
        return Promise.resolve(true);
    }
```

image

rajsite: I reset myself to look at it again. Both those examples seem convoluted under the assumption that we should not mix `async-await` style code and `Promise()/then()` style code in the same function. Neither should be encouraged and that particular example should not be `async`:

```ts
public asyncAwesomenessC(): Promise<boolean> {
    return Promise.resolve(true);
}
```

But I'd need to read through the error handling parts again to remember the context

rajsite: Looking at it again, I think I'd prefer the `always` configuration. Any Promise returned by an `async` function should be awaited. If you are making a helper, etc. that directly deals with `Promise` instances then you could choose to not mark the function async.

```ts
function timeout(timeMs: number): Promise<void> {
    return new Promise(resolve => setTimeout(resolve, timeMs));
}
```

If you end up in a situation where you mix and match `async` functions and `Promise` handling:
1. Don't, betting you can refactor the Promise creation into a separate function
2. Even if not, all promises will be awaited making the return paths easier to reason about

```ts
async function timeout(timeMs: number): Promise<void> {
    return await new Promise(resolve => setTimeout(resolve, timeMs));
}
```

I'm gonna try that change in nimble and see what happens

rajsite: This is the resulting PR in nimble: https://github.com/ni/nimble/pull/2412

I think it looks like an improvement and the fixer worked well. Fixer results needed a couple of tweaks to make some inferred generics explicit (comment in PR) but was a straightforward change.

Only difference is this thought:

> If you are making a helper, etc. that directly deals with Promise instances then you could choose to not mark the function async.

Was incorrect. We have [promise-function-async](https://typescript-eslint.io/rules/promise-function-async/) enabled which forces any function that returns a Promise to be async. But I actually like how that combination works. Every promise awaited and every function that returns a Promise is async.

packages/eslint-config-angular/requiring-type-checking.js
```diff
@@ -8,10 +8,9 @@ module.exports = {
             files: ['*.spec.ts'],
             rules: {
                 /*
-                    The jasminewd2 library used by Angular applications results in a significant number of
-                    floating promises and unbound methods so these rules are disabled for test specs in Angular projects.
+                    Spies used by Angular application tests result in a significant number of
+                    unbound methods so this rule is disabled for test specs in Angular projects.
                 */
-                '@typescript-eslint/no-floating-promises': 'off',
                 '@typescript-eslint/unbound-method': 'off',
```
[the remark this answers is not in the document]
TrevorKarjanis: Yes, this is very common, and the input to expect needs to be the spy, not a copy.

TrevorKarjanis (approved):

mure (approved):

tests/typescript-requiring-type-checking/index.ts
jattasNI: Thanks for trying it out! In addition to the reasons you mentioned, I like that `always` doesn't require different code in try-catch vs outside try-catch.

I updated the rule definition and made the test case pass. @rajsite once you approve I think this PR is good to complete!

PR #158, opened 2024-09-19
Disable explicit member accessibility for constructors

### LINKED ISSUE #100: Ignore constructors in @typescript-eslint/explicit-member-accessibility
@typescript-eslint/explicit-member-accessibility by default is forcing us to add `public` before every constructor, even though 99.9% of the time the implicit public is expected.
We could set `[ { overrides: { constructors: 'no-public' } } ]` to change this: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-member-accessibility.md#disallow-the-use-of-public-on-a-given-member

jattasNI:
# Rationale

Resolves #100. Constructors may now be marked `public`, `private`, or have no accessibility modifier. Previously they had to be marked `public` or `private`.

# Implementation

Set `overrides: { constructors: 'off' }` in the TS rule configuration.

# Testing

Added a constructor to the TypeScript test and verified the rule previously errored if it didn't have a modifier but now allows this.

### DIFF
--- a/change/@ni-eslint-config-typescript-b2c9d1d5-86de-4fd6-b66f-c5a696790e65.json
+++ b/change/@ni-eslint-config-typescript-b2c9d1d5-86de-4fd6-b66f-c5a696790e65.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Disable explicit-member-accessibility for ctors",
+  "packageName": "@ni/eslint-config-typescript",
+  "email": "jattasNI@users.noreply.github.com",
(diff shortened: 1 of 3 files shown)


rajsite (approved):

TrevorKarjanis (approved):

jattasNI:
@mure since you filed the original issue and the changes in this PR are trivial, I'm gonna assume you're ok with it and complete the PR today.

mure (approved):

Issue #157, opened 2024-09-19
Support Angular 18


jattasNI:
This issue tracks the work to support Angular 18.

In addition to the normal work, these issues should be resolved / unblocked during the upgrade:

- [x] #153 

I don't think #150 depends on this work but if hasn't been fixed yet then it should be grouped with this too.

#155 is a successor of this work but shouldn't happen at the same time.
#92 is a successor of this work as well.

m-akinc:
Removed the link to #155 from this issue. The ESLint 9 upgrade is gated by the Angular 18 upgrade, but it is substantial enough that we don't want to do them at the same time.

Issue #153, opened 2024-08-13
Uptake new angular-eslint runtime-localize rule


jattasNI:
**Describe the issue**

@m-akinc added a new rule to angular-eslint in https://github.com/angular-eslint/angular-eslint/pull/1898 which should help catch cases where strings are tagged with `$localize` in scopes that don't permit it, like file-level `const`s. This is now available in [@angular-eslint/eslint-plugin](https://www.npmjs.com/package/@angular-eslint/eslint-plugin/v/18.3.0). You can [see the docs here](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin/docs/rules/runtime-localize.md).

We should enable this rule by default in our ruleset.

jattasNI:
I tried to fix this in the [runtime-localize branch](https://github.com/ni/javascript-styleguide/tree/runtime-localize) but hit snags.
1. It requires installing `@angular-eslint/eslint-plugin@18.3.0` but just bumping the version results in dependency conflicts with `npm install`
2. I tried regenerating `package-lock.json` and that successfully installed the dependency but it seemed to break other rules. Our tests for `@angular-eslint/template/i18n` started failing claiming the content didn't have an `i18n` attribute even though it does.

I suspect both of these issues are because [the angular-eslint version is supposed to align with the Angular version](https://github.com/angular-eslint/angular-eslint?tab=readme-ov-file#supported-angular-cli-versions) but we're on Angular 17. If this is true then this issue needs to wait on the Angular 18 upgrade.

Issue #150, opened 2024-07-24
Upgrade @typescript-eslint/eslint-plugin @typescript-eslint/parser


TrevorKarjanis:
After upgrading to Angular 17, the following warning is received when running the linter.

```shell
WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
You may find that it works just fine, or you may not.
SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <5.2.0
YOUR TYPESCRIPT VERSION: 5.4.5
Please only submit bug reports when using the officially supported version.
```

This is because @typescript-eslint/eslint-plugin @typescript-eslint/parser are incompatible with TypeScript@5.4.5. This is not an npm dependency. [It is defined at runtime](https://github.com/typescript-eslint/typescript-eslint/blob/88b7463c89f22559facd5a00460bdc930e6eb1a2/packages/typescript-estree/src/parseSettings/warnAboutTSVersion.ts#L9) in @typescript-eslint/typescript-estree.

These packages, version `^5.62.0`, are peer dependencies of [@ni/eslint-config-typescript](https://github.com/ni/javascript-styleguide/blob/6d3cf459ec54310f6e0a01254e8d2d540e00d789/packages/eslint-config-typescript/package.json#L36), so upgrading them requires forcing the install. @typescript-eslint/typescript-estree@7.2 supports <5.5, so a minimal upgrade of @typescript-eslint/eslint-plugin@7.2 @typescript-eslint/parser@7.2 would be required.

Have we considered upgrading them, or why have they not been upgraded?

rajsite:
> Have we considered upgrading them, or why have they not been upgraded?

No specific reason I'm aware. Don't have a rennovate integration so would be manual. Looking at the getting started today they don't even pull from the scoped packages like we are which is interesting: https://typescript-eslint.io/getting-started/

jattasNI:
We agreed we should fix this soon. May need to do a bit of testing to ensure the new packages don't change the behavior of any rules in unexpected ways.

PR #160, opened 2025-01-07
Update to Angular 18

m-akinc:
Addresses #157 

In addition to updating the dependency version, I'm fixing the following:
- #150
- #153 

## Implementation
Updated dependencies:
 - `@angular/core` to `18.2.13` (latest <19)
 - `@angular-eslint` packages to `18.4.3` (latest <19)
 - `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` to `7.18.0` (latest <8)   
 - After updating the version of  `@typescript-eslint/parser`, I began getting the following test failure: 
     ```
     Error: Error while loading rule '@typescript-eslint/prefer-optional-chain': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
     Parser: C:\dev\javascript-styleguide\node_modules\@typescript-eslint\parser\dist\index.js
     Occurred while linting C:\dev\javascript-styleguide\tests\typescript\index.ts
     ```
     It turns out that `@typescript-eslint/prefer-optional-chain` now [does require type checking](https://typescript-eslint.io/rules/prefer-optional-chain), so I had to move it to the `requiring-type-checking.js` configuration.
  - `@typescript-eslint/eslint-plugin` version 6 [introduced significant changes](https://github.com/typescript-eslint/typescript-eslint/blob/9e8828b31fda6e0fa45e0397cd6cc4e1662ba356/packages/website/blog/2023-07-09-announcing-typescript-eslint-v6.md#user-facing-breaking-changes). In `requiring-type-checking.js`, I changed it from extending `@typescript-eslint/recommended-requiring-type-checking` to `@typescript-eslint/recommended-type-checked` and `@typescript-eslint/stylistic-type-checked`. Similarly, in `eslint-config-typescript/index.js`, I added `@typescript-eslint/stylistic` to the extended configs. Subsequently, the output of `npm run print-evaluated-rules:diff` showed over a dozen changes. I added overrides to minimize the scope of changes in this PR, and ultimately ended up with the following diff:
      -  `@typescript-eslint/consistent-generic-constructors` is new ("error") -- Despite adding an override to turn this rule off, it remains on. This may be a bug in `@typescript-eslint`. I don't think we want to let this block us, so clients should either fix reported violations or use in-source suppressions.
      - `@typescript-eslint/lines-between-class-members` was `["error","always",{"exceptAfterSingleLine":true}]`, now is `["error","always",{"exceptAfterSingleLine":true,"exceptAfterOverload":true}]` -- This seems like a harmless difference.
      - `no-new-native-nonconstructor` is new ("off") -- Explicitly turning off this new rule that defaulted to "error".
      - `@typescript-eslint/no-duplicate-enum-values` is new ("off") -- Explicitly turning off this new rule that defaulted to "error".
      - `@typescript-eslint/no-unsafe-declaration-merging` is new ("off") -- Explicitly turning off this new rule that defaulted to "error".
      - `@typescript-eslint/no-duplicate-type-constituents` is new ("off") -- Explicitly turning off this new rule that defaulted to "error".
      - `@typescript-eslint/no-redundant-type-constituents` is new ("off") -- Explicitly turning off this new rule that defaulted to "error".
      - `@typescript-eslint/no-unsafe-enum-comparison` is new ("off") -- Explicitly turning off this new rule that defaulted to "error".
      - `@angular-eslint/runtime-localize` is new ("error") -- We specifically want this enabled.

### Testing
- Ran `npm run print-evaluated-rules:diff` and handled differences as described above.
- Ran `npm run print-available-rules` with these changes:
   - `@typescript-eslint/no-duplicate-imports` - removed from TS Extensions
   - `@typescript-eslint/class-methods-use-this` - added to TS Extensions
   - `@typescript-eslint/max-params` - added to TS Extensions
   -  `@typescript-eslint/prefer-ts-expect-error` - moved from TS Recommended to Remaining TS Rules
   -  `@typescript-eslint/prefer-optional-chain` - moved from TS Recommended to TS Recommended (Requiring Type Checks)
   -  `@typescript-eslint/no-implicit-any-catch` - removed from Remaining TS Rules
   -  `@typescript-eslint/no-parameter-properties` - removed from Remaining TS Rules
   -  `@typescript-eslint/sort-type-union-intersection-members` - removed from Remaining TS Rules
   - `@typescript-eslint/no-empty-object-type` - added to Remaining TS Rules
   - `@typescript-eslint/no-unnecessary-parameter-property-assignment` - added to Remaining TS Rules
   - `@typescript-eslint/no-unsafe-function-type` - added to Remaining TS Rules
   - `@typescript-eslint/no-wrapper-object-types` - added to Remaining TS Rules
   - `@typescript-eslint/no-empty-object-type` - added to Remaining TS Rules
   - `@typescript-eslint/no-empty-object-type` - added to Remaining TS Rules
   - `@typescript-eslint/consistent-return` - added to TS Extensions (Requiring Type Checks)
   - `@typescript-eslint/only-throw-error` - added to TS Extensions (Requiring Type Checks)
   - `@typescript-eslint/prefer-destructuring` - added to TS Extensions (Requiring Type Checks)
   - `@typescript-eslint/prefer-promise-reject-errors` - added to TS Extensions (Requiring Type Checks)
   - `@typescript-eslint/no-array-delete` - added to TS Recommended (Requiring Type Checks)
   - `@typescript-eslint/no-confusing-void-expression` - moved from Remaining TS Rules (Requiring Type Checks) to TS Recommended (Requiring Type Checks)
   - `@typescript-eslint/no-duplicate-type-constituents` - moved from Remaining TS Rules (Requiring Type Checks) to TS Recommended (Requiring Type Checks)
   - `@typescript-eslint/no-redundant-type-constituents` - moved from Remaining TS Rules (Requiring Type Checks) to TS Recommended (Requiring Type Checks)
   - `@typescript-eslint/no-unnecessary-template-expression` - added to TS Recommended Requiring Type Checks
   - `@typescript-eslint/use-unknown-in-catch-callback-variable` - added to TS Recommended Requiring Type Checks
   - `@typescript-eslint/no-unnecessary-type-parameters` - added to Remaining TS Rules (Requiring Type Checks)
   - `@typescript-eslint/no-unsafe-unary-minus` - added to Remaining TS Rules (Requiring Type Checks)
   - `@typescript-eslint/no-useless-template-literals` - added to Remaining TS Rules (Requiring Type Checks)
   - `@typescript-eslint/prefer-find` - added to Remaining TS Rules (Requiring Type Checks)
   - `@angular-eslint/no-host-metadata-property` - moved from Angular Recommended to Remaining Angular
   - `@angular-eslint/runtime-localize` - added to Remaining Angular

### DIFF
--- a/change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
+++ b/change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
@@ -1,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Update to Angular 18",
+  "packageName": "@ni/eslint-config-angular",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/eslint-config-typescript/tools/print-available-rules.js
+++ b/packages/eslint-config-typescript/tools/print-available-rules.js
@@ -1,4 +1,4 @@
-const plugin = require('@typescript-eslint/eslint-plugin');
+const plugin = require('@typescript-eslint/eslint-plugin/dist');
 
 const isTrue = val => val !== undefined && val !== false;
 const recommended = key => isTrue(plugin.rules[key].meta.docs.recommended);
--- a/packages/eslint-config-typescript/tools/print-available-rules.js
+++ b/packages/eslint-config-typescript/tools/print-available-rules.js
@@ -1,4 +1,9 @@
-const plugin = require('@typescript-eslint/eslint-plugin/dist');
+// eslint-plugin-import doesn't know how to resolve entry points in commonjs modules that are
+// declared using "export" rather than the old "main" syntax in package.json.
+// See https://github.com/typescript-eslint/typescript-eslint/issues/7565
+//  and https://github.com/import-js/eslint-plugin-import/issues/2703
+// eslint-disable-next-line import/no-unresolved
+const plugin = require('@typescript-eslint/eslint-plugin');
 
 const isTrue = val => val !== undefined && val !== false;
 const recommended = key => isTrue(plugin.rules[key].meta.docs.recommended);
--- a/tests/playwright/.eslintrc.js
+++ b/tests/playwright/.eslintrc.js
@@ -1,13 +1,11 @@
 module.exports = {
+    extends: '@ni/eslint-config-playwright',
     root: true,
-    ignorePatterns: ['*.js'],
     overrides: [
         {
-            extends: '@ni/eslint-config-playwright',
-            files: ['*.ts'],
-            parserOptions: {
-                project: './tsconfig.json'
-            },
+            files: [
+                '*.ts',
+            ],
             rules: {
                 // In this example app, all files are allowed to refer to playwright packages from devDependencies
                 'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
--- a/tests/playwright/.eslintrc.js
+++ b/tests/playwright/.eslintrc.js
@@ -1,11 +1,13 @@
 module.exports = {
-    extends: '@ni/eslint-config-playwright',
     root: true,
+    ignorePatterns: ['*.js'],
     overrides: [
         {
-            files: [
-                '*.ts',
-            ],
+            extends: '@ni/eslint-config-playwright',
+            files: ['*.ts'],
+            parserOptions: {
+                project: './tsconfig.json'
+            },
             rules: {
                 // In this example app, all files are allowed to refer to playwright packages from devDependencies
                 'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,4 +4,4 @@
 # More info: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners
 
 # These owners will be the default owners for everything in the repo.
-* @jattasNI @mure @rajsite @TrevorKarjanis
+* @jattasNI @rajsite @TrevorKarjanis
--- a/change/@ni-eslint-config-angular-9ac0e40a-7052-4b98-a74f-2b443d4cb1d7.json
+++ b/change/@ni-eslint-config-angular-9ac0e40a-7052-4b98-a74f-2b443d4cb1d7.json
@@ -1,7 +1,0 @@
-{
-  "type": "patch",
-  "comment": "Exempt groupableParentColumnId from i18n rule",
-  "packageName": "@ni/eslint-config-angular",
-  "email": "7282195+m-akinc@users.noreply.github.com",
-  "dependentChangeType": "patch"
-}
--- a/change/@ni-eslint-config-typescript-c8f0224a-ee12-4ed6-af5b-d65afda76daf.json
+++ b/change/@ni-eslint-config-typescript-c8f0224a-ee12-4ed6-af5b-d65afda76daf.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update version of @typescript-eslint/parser and @typescript-eslint/eslint-plugin",
+  "packageName": "@ni/eslint-config-typescript",
+  "email": "7282195+m-akinc@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,35 +38,37 @@
       }
     },
     "node_modules/@angular-devkit/architect": {
-      "version": "0.1703.7",
-      "resolved": "https://registry.npmjs.org/@angular-devkit/architect/-/architect-0.1703.7.tgz",
-      "integrity": "sha512-SwXbdsZqEE3JtvujCLChAii+FA20d1931VDjDYffrGWdQEViTBAr4NKtDr/kOv8KkgiL3fhGibPnRNUHTeAMtg==",
+      "version": "0.1802.12",
+      "resolved": "https://registry.npmjs.org/@angular-devkit/architect/-/architect-0.1802.12.tgz",
+      "integrity": "sha512-bepVb2/GtJppYKaeW8yTGE6egmoWZ7zagFDsmBdbF+BYp+HmeoPsclARcdryBPVq68zedyTRdvhWSUTbw1AYuw==",
+      "license": "MIT",
       "peer": true,
       "dependencies": {
-        "@angular-devkit/core": "17.3.7",
+        "@angular-devkit/core": "18.2.12",
         "rxjs": "7.8.1"
       },
       "engines": {
-        "node": "^18.13.0 || >=20.9.0",
+        "node": "^18.19.1 || ^20.11.1 || >=22.0.0",
         "npm": "^6.11.0 || ^7.5.6 || >=8.0.0",
         "yarn": ">= 1.13.0"
       }
     },
     "node_modules/@angular-devkit/core": {
-      "version": "17.3.7",
-      "resolved": "https://registry.npmjs.org/@angular-devkit/core/-/core-17.3.7.tgz",
-      "integrity": "sha512-qpZ7BShyqS/Jqld36E7kL02cyb2pjn1Az1p9439SbP8nsvJgYlsyjwYK2Kmcn/Wi+TZGIKxkqxgBBw9vqGgeJw==",
+      "version": "18.2.12",
+      "resolved": "https://registry.npmjs.org/@angular-devkit/core/-/core-18.2.12.tgz",
+      "integrity": "sha512-NtB6ypsaDyPE6/fqWOdfTmACs+yK5RqfH5tStEzWFeeDsIEDYKsJ06ypuRep7qTjYus5Rmttk0Ds+cFgz8JdUQ==",
+      "license": "MIT",
       "peer": true,
       "dependencies": {
-        "ajv": "8.12.0",
-        "ajv-formats": "2.1.1",
-        "jsonc-parser": "3.2.1",
-        "picomatch": "4.0.1",
+        "ajv": "8.17.1",
+        "ajv-formats": "3.0.1",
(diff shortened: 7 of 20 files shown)


change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
m-akinc, line 1: Won't submit until ready to update clients.

change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
jattasNI, line 2: >Ran npm run print-available-rules and the only new, unset rule available in the updated version is @angular-eslint/runtime-localize (which we implemented), and that can be turned on as needed.

I would like to turn on the rule by default in the styleguide and to consider that within scope of the Angular upgrade feature if possible. We could either turn it on in this PR or in a follow up. Either way we'd want to communicate the change at UI WG so that client teams can fix violations (or, less ideally, temporarily disable the rule).

If you don't think you'll get to it in this feature, we should discuss how to make it happen afterward.

change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
```diff
@@ -0,0 +1,7 @@
...
+  "comment": "Update to Angular 18",
+  "packageName": "@ni/eslint-config-angular",
```
jattasNI, line 4: FYI, #157 exists to track this work. It links to a few related issues: could you evaluate fixing at the same time?

change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: >Ran npm run print-evaluated-rules:diff and the only change is that the no-host-metadata-property rule was removed from the recommended eslint-plugin config (rule is deprecated in version 18 and fully removed in version 19). This should not break clients, as the rule still exists (so clients can still turn it on/off) and the default config is just more permissive.

I'd like to figure out the long term plan for adapting to this deprecation and if there's any work to do, decide whether to do it now or in the 19 upgrade.

Here is the PR that removed that rule: https://github.com/angular-eslint/angular-eslint/pull/2113 

It looks like Angular changed their recommendation to prefer this syntax
```
@Component({
  selector: '[mat-menu-item]',
  host: {
    '[class.mat-menu-item-highlighted]': '_highlighted',
  }
})
```

over the `@HostListener` and `@HostBinding` directives:

```
@HostBinding('class.my-menu-item-highlighted')
  _highlighted: boolean = false;
```

[This reddit post](https://www.reddit.com/r/Angular2/comments/feoq5r/change_my_mind_the_host_metadata_property_is_way/?rdt=44563) has some good examples of before / after.

They have an unimplemented feature request to write the rule to enforce that new preference: https://github.com/angular-eslint/angular-eslint/issues/2189

I searched our codebase and didn't see anything setting `no-host-metadata-property` (except for [one non-maintained codebase](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=no-host-metadata-property&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//Web/Packages/TSLintConfigSystemLink/tslint.json)) but I did see several instances of [`@HostListener`](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=HostListener&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//Web/Workspaces/AssetManager/src/app/app.component.ts) and [`@HostBinding`](https://dev.azure.com/ni/DevCentral/_search?action=contents&text=HostBinding&type=code&lp=code-Project&filters=ProjectFilters%7BDevCentral%7DRepositoryFilters%7BSkyline%7D&pageSize=25&result=DefaultCollection/DevCentral/Skyline/GBmaster//Web/Workspaces/AssetManager/src/app/app.component.ts). So I think we could either:
1. keep this PR as-is and nothing will break, but we'll be violating Angular's new recommendation for `@HostListener`/`@HostBinding`. 
2. do some work to migrate / delete those usages. That work could happen now or during the 19 upgrade or whenever that new rule is introduced

Any votes between 1 and 2?

jattasNI (changes requested):

change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
rajsite: One of the linked issues is migrating to eslint 9 which I think we should do in a follow-up instead after the Angular 18 upgrade. I suspect and commented on https://github.com/ni/javascript-styleguide/issues/155https://github.com/ni/javascript-styleguide/issues/155 that eslint 9 migration probably will take some effort.

change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
m-akinc: I've turned the rule on in eslint-config-angular. We can make fixing any issues in clients part of the work to migrate them to Angular 18.

change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
m-akinc: Treating the ESLint 9 update as a separate task, but I'll link/fix the other issues with this PR.

change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
m-akinc: I vote 1. I'm not sure the [rationale for the reversed convention](https://github.com/angular/angular/discussions/49682#discussioncomment-5515138) is compelling enough to warrant taking any action now. If there were a rule to enforce the new, reversed recommendation, I'd be more convinced that we should turn it on and fix violations.

packages/eslint-config-typescript/tools/print-available-rules.js
```diff
@@ -1,4 +1,4 @@
-const plugin = require('@typescript-eslint/eslint-plugin');
+const plugin = require('@typescript-eslint/eslint-plugin/dist');
```
m-akinc, line 1: The new version of `eslint-plugin`'s `package.json` file no longer has `"main": "dist/index.js",` and explicitly declares `"type": "commonjs"`. I don't know why it changed, but now we need to specify `/dist` or we get a lint error about the module not being found.

packages/eslint-config-typescript/tools/print-available-rules.js
```diff
@@ -1,3 +1,8 @@
+// eslint-plugin-import doesn't know how to resolve entry points in commonjs modules that are
+// declared using "export" rather than the old "main" syntax in package.json.
+// See https://github.com/typescript-eslint/typescript-eslint/issues/7565
+//  and https://github.com/import-js/eslint-plugin-import/issues/2703
+// eslint-disable-next-line import/no-unresolved
```
m-akinc, line 5: The new version of `eslint-plugin`'s `package.json` file no longer has `"main": "dist/index.js"` and instead uses the newer `"exports"` property. Unfortunately, `eslint-plugin-import` doesn't know how to resolve entry points in commonjs modules that are declared using `"export"`. See https://github.com/typescript-eslint/typescript-eslint/issues/7565  and https://github.com/import-js/eslint-plugin-import/issues/2703

change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
jattasNI: I'm ok with 1 too. This doesn't seem very common in or code or very consequential if we do it the "wrong" way.

jattasNI (commented):
Looks good to me, but withholding final approval until we're ready to start the release train.

tests/playwright/.eslintrc.js
```diff
@@ -1,11 +1,13 @@
...
+    ignorePatterns: ['*.js'],
     overrides: [
         {
-            files: [
-                '*.ts',
-            ],
+            extends: '@ni/eslint-config-playwright',
+            files: ['*.ts'],
+            parserOptions: {
+                project: './tsconfig.json'
```
rajsite, line 9: > In any case, clients should already be setting parserOptions.project for running eslint on their ts files, so I wouldn't expect this to cause issues.

That was not mandatory before and clients had the option of adopting rules that did not need a tsconfig for type checking (with the trade-off of supporting a smaller rule set). These configurations currently don't require tsconfigs, and the `requiring-type-checking` rulesets are the ones that should require tsconfigs.

I'd like to not couple that change with this PR and not change the behavior of the typescript vs typescript-requiring-type-checking. If that is not possible then we should open the discussion to merging the rulesets since the distinction would no longer be meaningful.

m-akinc: It turns out that rule now does in fact require type checking, so I moved it to the other config. I also discovered that there were a lot more changes than I had originally thought (I must not have done a `npm i` in between running `print-evaluated-rules`). I've updated the PR description with detailed information about how I handled those changes.

rajsite (approved):

change/@ni-eslint-config-angular-718354fd-7ab1-4395-ac53-e8505b9622bf.json
m-akinc: Ready to go.

PR #161, opened 2025-01-23
Remove @mure as a code owner

mure:
I'm leaving NI and will no longer be a code owner.


rajsite (approved):
"accidentally" removed Jesse 🕵️‍♂️, but fixed that up 😆

mure:
Oof, good catch. Guess my brain is already on vacation 😆

Issue #148, opened 2024-06-26
Stop providing configuration for deprecated formatting rules


m-akinc:
Formatting rules were [deprecated in ESLint 8.53.0](https://eslint.org/blog/2023/10/deprecating-formatting-rules/). As suggested in the linked document, we should probably replace these with their counterparts from `@stylistic/eslint-plugin-js(ts)`.

jattasNI:
Consensus among our working group is to migrate to the new `stylistic` packages in order to get off of deprecated rules. But we have until ESLint v10 to do so since that's when the deprecated rules will be deleted.

Longer term we could look at using one of the formatters in the linked blog post instead. Probably not prettier since it isn't very fast or configurable, but the [dprint](https://dprint.dev/) one looks worth exploring.

We don't anticipate major client impact. If any clients were overriding any of the now deprecated rules, they may need to do a bit of work to uptake this migration. We should inspect a couple client projects to see if that's the case, but it doesn't seem likely.

This is available to work on when someone has bandwidth.

jattasNI:
I scanned Skyline for usages of the deprecated rules. The only instances I found were:

- two projects configure `eol-last: error`
- one inline disable of `function-paren-newline`
- one inline disable and one project-wide configuration of `indent`
- one project disables `lines-between-class-members`
- one project unnecesarily sets `no-trailing-spaces: error`
- a few inline disables of `object-curly-newline`
- one inline disable of quote-props
- one override of `spaced-comment`

In summary: nothing that should stop us from doing this work but a handful of places that might need to be looked at when uptaking this change if the rule names or configurations change.

Issue #116, opened 2022-12-01
Remove dependency on Airbnb ruleset


jattasNI:
We think it would be a good idea to remove the dependency on the Airbnb ruleset since it makes it more complicated to understand how rules are configured in this repo.

PR #167, opened 2025-09-01
Remove Airbnb ruleset dependency from style guide

gokulprasanth-ni:
# **Justification**

- [Remove dependency on Airbnb ruleset](https://github.com/ni/javascript-styleguide/issues/116)

Note: [Stop providing configuration for deprecated formatting rules](https://github.com/ni/javascript-styleguide/issues/148) will be taken as follow-up PR.

# **Implementation**

- Removed the Airbnb dependency.
- Copied the rules folder along with the same folder structure from [eslint-config-airbnb-base](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/index.js), since airbnb-base internally referenced it.
- Distributed rules from index.js into their relevant rule files, now that the sources are in our control.

# **Testing**

- Built locally and integrated with our app. Ran linting, which completed successfully.
image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,4 +64,4 @@ The following are guidelines for when to release as `patch`, `minor`, or `major`
 2. Make a `minor` release for new rules or changes to rules.
 
    This results in most updates to the lint rules resulting in breakages during continuous integration and requiring source changes despite a `minor` version change. This was deemed acceptable as projects are expected to use `package-lock.json` files so updates are intentional and we expect updates to be infrequent.
-3. Make a `major` release when changes to dependencies are performed such as requiring a newer Node.js version, or an ESlint version update.
+3. Make a `major` release when changes to dependencies are performed such as requiring a newer Node.js version or an ESlint version update.


change/@ni-eslint-config-typescript-67da7439-6c61-46c1-9440-60866dfdaae9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
vivinkrishna-ni, line 2:
```suggestion
  "type": "none",
```

I think removing comments alone is not considered a `patch` version bump. Please update the `dependentChangeType` to `none` as well if you agree.

change/@ni-eslint-config-javascript-782dc014-2db5-4c39-94e6-ad68dd2b3470.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
vivinkrishna-ni, line 2: Ensuring the version as per the [COTRIBUTING.md doc](https://github.com/ni/javascript-styleguide/blob/main/CONTRIBUTING.md#rules-for-version-change) that updates to peer dependency is considered a `major` version bump 👍

change/@ni-eslint-config-typescript-67da7439-6c61-46c1-9440-60866dfdaae9.json
```diff
@@ -0,0 +1,7 @@
+{
```
vivinkrishna-ni, line 1: Should we call the change of copying the ruleset of Airbnb into the [CONTRIBUTING.md doc](https://github.com/ni/javascript-styleguide/blob/d751343c425fe27e29ec925812a5c91ba7a5330b/CONTRIBUTING.md?plain=1#L26C58-L26C72) referencing its dependency?

Open for suggestions from other reviewers.

packages/eslint-config-javascript/index.js
```diff
@@ -1,460 +1,16 @@
...
-            C#, and Python).
-        */
-        'space-before-function-paren': ['error', {
-            anonymous: 'always',
-            named: 'never',
-            asyncArrow: 'always'
-        }],
-    },
-};
+        './rules/best-practices',
```
vivinkrishna-ni, line 3: Would it be better to add a generic code comment here, mentioning that some of the rules are adapted from Airbnb's style guide, and include a link to the codebase? We could also note that certain values are overridden in specific rule files based on NI recommendations.

vivinkrishna-ni (approved):

jattasNI (commented):
Just some very preliminary feedback. I still need to wrap my head around the rules changes.

change/@ni-eslint-config-typescript-67da7439-6c61-46c1-9440-60866dfdaae9.json
jattasNI: I'm actually not certain which configuration wins if a library specifies `"type": "none"` but a dependency specifies `"dependentChangeType": "major"` (as we do in the eslint-config-javascript changefile). But I would vote to make this a major change to all libraries so think it would be most clear to make this changefile more closely match the javascript one: use `major` change type and set the comment to `"Remove Airbnb dependency from style guide"`

packages/eslint-config-javascript/index.js
jattasNI: @rajsite @TrevorKarjanis I'd like your opinion on how we want to lay out our rules and how much of a link we want back to Airbnb in comments / history.

1. This PR's proposed approach: copy Airbnb's file layout (index.js that imports several rules files). Then move our existing rules from index.js into the corresponding rules file, overwriting the Airbnb rule.
   - Pros:
     - file layout is organized so no rules file is very big
   - Cons:
     - lose history of how we made rules decisions (links to NI PRs in blame history and notes about which came from Airbnb)
     - when new eslint rules are introduced we have to figure out which rules file they belong in

2. Maintain current structure: single index file containing all the rules. We could copy all the airbnb rules (except the ones we override) into that file, maintaining the alphabetical ordering.
   - Pros: 
       - git blame would show whether rules came from historic NI rule debates or Airbnb (this PR)
       - no need to debate which file to put new rules in, it's just alphabetical in index.js
   - Cons:
       - probably more copy-paste work to intersperse all those rules in one file (though hopefully genai can help)
       - index.js will be a huge file, probably 1000-2000 lines

I rejected other alternatives like these because I don't want to retain Airbnb cruft after this PR, I want to sever the link completely.
3. Add a comment to each rule annotating whether it came from NI or Airbnb
4. Keep the Airbnb rules in separate files from the NI overrides forever

I have a slight preference for 2 but interested in other opinions or ideas or assessments of the cost.

rajsite: I think I'm ultimately somewhere in the middle but leaning closer to option 1. I don't think preserving the history on the JS package is as interesting as the typescript / angular packages. I like that the airbnb rules for es6, imports, and node are split based on the additional env / parserOptions / plugins they depend on but I don't really care for the other categorizations of  best-practices, errors, styles, variables.

I think I lean to 1 today and then create a separate tech debt / task to do the same kind of env/plugin based split across packages. i.e. the typescript rules should have a separate imports.js configuration based on the imports plugin as well. That task would also collapse the base rules down together from the files with arbitrary categories.

If there is an alignment on that (use categorizations based on env / plugins and remove the arbitrary / invented categorizations) then I'd propose we stick with 1 today and save that other realignment for a follow-up. In that world I'd probably say as we add the stylistic plugin in the next PR for https://github.com/ni/javascript-styleguide/issues/148 we go ahead and put that config based on that plugin in it's own file.

jattasNI: Yeah, I can see that there's value in grouping rules in separate files if we can come up with groupings that make it obvious where any given rule should go. And I agree that different environments and plugins in separate files is a sane organization.

So I'd be ok with either:
1. the current PR organization plus filing tech debt to revisit 
2. combining all the rules files that don't include environment configuration into one file (`base.js`?) in this PR

rajsite: I'd prefer 1 for today so we can see where things settle after the eslint 9 transition and then re-organize.

change/@ni-eslint-config-typescript-67da7439-6c61-46c1-9440-60866dfdaae9.json
vivinkrishna-ni: > Are you asking what we should do about this last search result in the repo for "airbnb"?

Yes, I believe we can proceed with updating the document according to your suggestion.

gokulprasanth-ni: Updated as per the suggestion

change/@ni-eslint-config-javascript-782dc014-2db5-4c39-94e6-ad68dd2b3470.json
gokulprasanth-ni: We are fine with type: "path" for this change. Waiting for jesse's thoughts too.

change/@ni-eslint-config-typescript-67da7439-6c61-46c1-9440-60866dfdaae9.json
gokulprasanth-ni: I will update this based on the decision taken from the other thread.

gokulprasanth-ni: Updated the type and dependentChangeType as "patch.

packages/eslint-config-javascript/index.js
gokulprasanth-ni: Created an issue https://github.com/ni/javascript-styleguide/issues/168 for tracking this.

CONTRIBUTING.md
```diff
@@ -23,7 +23,7 @@ The TypeScript extension rules (which are rules that extend the behavior of exis
 
 ### Checking the evaluated rules
```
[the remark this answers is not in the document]
gokulprasanth-ni: Most of the newly added rules are set to `off`. The only notable change is `no-spaced-func`. It was `error` in `eslint-config-airbnb-base v15.0.0` but is now `off` in the newer version. If we still want to enforce that spacing style, we can update it in the upcoming PR of replacing the deprecated rules as this rule is deprecated now. For `spaced-comment`, our override simplifies the rule to just `error` (removing Airbnb’s extra exceptions/markers); earlier it looked like the override wasn’t taking effect, but it is now working as intended.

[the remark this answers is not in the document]
gokulprasanth-ni: I have updated the `spaced-comment` rule value as suggested. Regarding the `disallowRedundantWrapping` option for the `prefer-regex-literals` rule, its default value is `false`, which does not match our existing configuration. Therefore, I have explicitly set this option to `true` to ensure the rule continues to function as before and everything works as expected.

TrevorKarjanis:
@rajsite @jattasNI These PR mentions hit a different busy inbox, and I have been busy. Nice merge.

PR #170, opened 2025-09-15
Update javascript-styleguide packages to be compatible with eslint v9

### LINKED ISSUE #155: Upgrade to ESLint 9
**Describe the issue**

[ESLint 8 is end of life on October 5, 2024](https://eslint.org/blog/2024/09/eslint-v8-eol-version-support/). `@ni/eslint-config-javascript` [currently depends on `"eslint": "^8.57.0",`](https://github.com/ni/javascript-styleguide/blob/5fcd774b7af469b88853a7aba64e0a92af897632/packages/eslint-config-javascript/package.json#L33). There is [a v9 migration guide](https://eslint.org/docs/latest/use/migrate-to-9.0.0) that lists several breaking changes. From a quick scan:
1. config files are being renamed from eslintrc to eslint.config.js but there's a way to use the deprecated format
2. eslint:recommended was updated
3. several rule-specific configuration changes

Related: #150

gokulprasanth-ni:
# Justification

- #155

Resolves #155

# Implementation

- Remove all the legacy config files and replace them with flat config files.
- Update the configurations to use [flat config format](https://eslint.org/docs/latest/use/configure/migration-guide)
- Update the dependencies version that support eslint v9.
- Update the READMe doc related to this eslint v9 change.

# Testing

- Packed and applied the updated style guide in our client app - linting passed successfully.
image


tests/angular/eslint.config.js
```diff
@@ -0,0 +1,59 @@
...
+import angular from '@ni/eslint-config-angular';
+import angularTemplate from '@ni/eslint-config-angular/template';
+import angularRequiringTypeChecking from '@ni/eslint-config-angular/requiring-type-checking';
+import { ignoreAttributes } from '@ni/eslint-config-angular/template/options';
+
+import typescriptPlugin from 'typescript-eslint';
+import angularTemplatePlugin from 'angular-eslint';
+import { defineConfig } from 'eslint/config';
+
+const tsconfigRootDir = path.dirname(fileURLToPath(import.meta.url));
```
vivinkrishna-ni, line 13:
```suggestion
const tsConfigRootDir = path.dirname(fileURLToPath(import.meta.url));
```
Should this be renamed?

tests/playwright-requiring-type-checking/eslint.config.js
```diff
@@ -0,0 +1,33 @@
+import { fileURLToPath } from 'url';
+import path from 'path';
+import playwright from '@ni/eslint-config-playwright';
+import playwrightRequiringTypeChecking from '@ni/eslint-config-playwright/requiring-type-checking';
+
+const tsconfigRootDir = path.dirname(fileURLToPath(import.meta.url));
```
vivinkrishna-ni, line 6:
```suggestion
const tsConfigRootDir = path.dirname(fileURLToPath(import.meta.url));
```

tests/print-evaluated-rules/index.js
```diff
@@ -73,28 +91,53 @@ const { hideBin } = require('yargs/helpers');
...
+        }, {});
+
+    const angularDivergedRules = getDivergedRules(
+        angularRules.map(cfg => cfg.rules)
+    );
+    const angularTemplateDivergedRules = getDivergedRules(
+        angularTemplateRules.map(cfg => cfg.rules)
+    );
+    const angularHasDivergedRules = !!Object.keys(angularDivergedRules).length
+      || !!Object.keys(angularTemplateDivergedRules).length;
```
vivinkrishna-ni, line 116:
```suggestion
        || !!Object.keys(angularTemplateDivergedRules).length;
```
nit

tests/print-evaluated-rules/index.js
```diff
@@ -73,28 +91,53 @@ const { hideBin } = require('yargs/helpers');
...
-            .filter(key => !configTypescript.rules[key] && !configTypescriptTypechecked.rules[key])
-            .reduce((config, key) => {
-                config[key] = rules[key];
-                return config;
-            }, {})
+    const getDivergedRules = rules => Object.keys(rules)
+        .filter(key => !key.startsWith('@angular-eslint'))
+        .filter(
+            key => !configTypescript.rules[key]
+              && !configTypescriptTypechecked.rules[key]
```
vivinkrishna-ni, line 102:
```suggestion
                && !configTypescriptTypechecked.rules[key]
```
nit

tests/print-evaluated-rules/index.js
```diff
@@ -73,28 +91,53 @@ const { hideBin } = require('yargs/helpers');
...
+        )
+        .reduce((config, key) => {
+            config[key] = rules[key];
+            return config;
+        }, {});
+
+    const angularDivergedRules = getDivergedRules(
+        angularRules.map(cfg => cfg.rules)
```
vivinkrishna-ni, line 110:
```suggestion
        angularRules.map(config => config.rules)
```
nit

package.json
```diff
@@ -51,6 +51,6 @@
     "beachball": "^2.41.0"
   },
   "peerDependencies": {
-    "typescript": "~4.8.2"
+    "typescript": "^5.0.0"
```
vivinkrishna-ni, line 54:
```suggestion
    "typescript": "~5.0.0"
```
Should we stick to `~` where only the patch version is accepted? Since this is a JavaScript style guide that other projects depend on for consistent linting behavior, I think we might need to use `~` instead of `^` for dependencies like TypeScript and ESLint.

README.md
```diff
@@ -69,80 +69,102 @@ After installing the lint configuration packages, follow the configuration instr
 
 ### JavaScript configuration
 
-Extend `@ni/eslint-config-javascript` in the [ESLint configuration](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats).
+Spread `@ni/eslint-config-javascript` configurations in your [ESLint flat config](https://eslint.org/docs/latest/use/configure/configuration-files-new) (`eslint.config.js`):
```
vivinkrishna-ni, line 72:
```suggestion
Spread `@ni/eslint-config-javascript` configurations in your [ESLint flat configuration](https://eslint.org/docs/latest/use/configure/configuration-files-new) (`eslint.config.js`):
```
Applicable to other descriptions as well.

change/@ni-eslint-config-angular-2b894e10-2f6e-4ff2-ade2-fae0335c38d8.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
+  "comment": "Update eslint-config-angular to be compatible with eslint v9",
+  "packageName": "@ni/eslint-config-angular",
+  "email": "gokulprasanth.ravi@emerson.com",
```
vivinkrishna-ni, line 5: If you'd like to keep the email address private, enable it from the [settings](https://github.com/settings/emails).

vivinkrishna-ni (approved):

packages/eslint-config-angular/index.js
```diff
@@ -1,139 +1,134 @@
...
-        'plugin:@angular-eslint/template/process-inline-templates',
-        '@ni/eslint-config-typescript'
-    ],
-    rules: {
-        /*
-            Overrides to Angular recommended rules:
-            https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.json
-        */
-
-        '@angular-eslint/no-host-metadata-property': 'error',
```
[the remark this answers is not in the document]
gokulprasanth-ni: The angular-eslint v19 has declared this rule as deprecated and asked users to use host meta property instead of @HostListener(). But since we are using these host decorators in the NI repositories I have used the angular-eslint v18.4.3 itself. This version is also compatible with eslint v9.

packages/eslint-config-typescript/tools/print-available-rules.js
```diff
@@ -1,22 +1,23 @@
+/* eslint-disable no-undef */
```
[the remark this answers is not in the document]
gokulprasanth-ni: It is not needed. I missed to add node globals. I have updated it.

change/@ni-eslint-config-angular-2b894e10-2f6e-4ff2-ade2-fae0335c38d8.json
```diff
@@ -0,0 +1,7 @@
+{
```
jattasNI, line 1: Some manual testing that I did:
1. verified that introducing a lint error in each of our test files still results in failing that test 
2. verified that vscode lint annotations and fix commands still work (required restarting vscode after switching to this branch and installing dependencies) 

👍

packages/eslint-config-javascript/eslint.config.js
```diff
@@ -0,0 +1,13 @@
+import javascript from '@ni/eslint-config-javascript';
+
+export default [
+    ...javascript,
+    {
+        files: ['**/*.js'],
+        rules: {
```
jattasNI, line 7: It looks like these are new rule configurations for this project's source, is that correct? Can you comment why they were needed? Assuming there's a reasonable justification, i general it's nice to add a short code comment justifying deviations from the standard ruleset.

Similar question for other packages too.

packages/eslint-config-javascript/rules/imports.js
```diff
@@ -1,15 +1,4 @@
...
-    },
-    parserOptions: {
-        ecmaVersion: 6,
-        sourceType: 'module'
-    },
-    plugins: [
-        'import'
-    ],
-
+export const imports = {
```
jattasNI, line 1: Can you comment on the deletion of this configuration? Is there no longer a way to configure these settings? Does this result in a behavior change?

```js
    env: {
        es6: true
    },
    parserOptions: {
        ecmaVersion: 6,
        sourceType: 'module'
    },
    plugins: [
        'import'
    ],
```

packages/eslint-config-javascript/rules/style.js
```diff
@@ -465,7 +459,7 @@ module.exports = {
             deprecated in favor of func-call-spacing
```
jattasNI, line 459: Good catch on this re-configuration bug. Can you bring the comment over too and replace this comment?

packages/eslint-config-playwright/index.js
```diff
@@ -1,82 +1,85 @@
...
+import playwright from 'eslint-plugin-playwright';
 
-        /*
-            Playwright marks this as recommended but configures it as 'warn'.
-            Deeply nested describe blocks are likely an indication that test files are too complicated.
-        */
-        'playwright/max-nested-describe': 'error',
+export default [
+    ...typescript,
+    playwright.configs['flat/recommended'],
```
jattasNI, line 6: Looks like we switched from `'plugin:playwright/playwright-test'` to `playwright.configs['flat/recommended'],`. Have you verified that they're equivalent? Or is this a behavior change?

packages/eslint-config-typescript/index.js
```diff
@@ -1,226 +1,233 @@
...
+            /*
+                Overrides to TypeScript recommended rules:
+                https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.ts
+            */
+
+            '@typescript-eslint/explicit-module-boundary-types': 'error',
+
+            /*
+                [strict-null-checks]
+                With `stickNullChecks` disabled, non-null assertions are discouraged (this rule is configured as 'error')
```
jattasNI, line 40:
```suggestion
                With `strictNullChecks` disabled, non-null assertions are discouraged (this rule is configured as 'error')
```

packages/eslint-config-typescript/index.js
```diff
@@ -1,226 +1,233 @@
...
+            */
+            '@typescript-eslint/no-non-null-assertion': 'error',
+
+            'valid-typeof': 'off',
+
+            /*
+                Overrides to TypeScript stylistic rules:
+                https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/stylistic.ts
+            */
+
```
jattasNI, line 53: I didn't notice this in the previous PR but it's weird to have an empty section. Could you either delete this comment or change it to say that there aren't any overrides?

[Edit]Actually, it looks like there are some stylistic overrides below. Can you move this down there?

packages/eslint-config-typescript/lib/extensions-requiring-type-checking.js
```diff
@@ -7,12 +7,12 @@ module.exports = {
         '@typescript-eslint/no-implied-eval': 'error',
 
         'no-throw-literal': 'off',
-        '@typescript-eslint/no-throw-literal': 'error',
+        '@typescript-eslint/only-throw-error': 'error',
```
jattasNI, line 10: Can you comment on this rule change? What was the rationale? Is it a behavior change? If it is, it would be good to call it out explicitly in the changefile.

change/@ni-eslint-config-angular-2b894e10-2f6e-4ff2-ade2-fae0335c38d8.json
```diff
@@ -0,0 +1,7 @@
...
+  "type": "major",
```
rajsite, line 2: Since this is a major change and we are inventing our new api as part of it (the names / conventions for exports) I want to concretely propose the following three changes:

1. We use named exports for the configuration instead of default exports.
   - The javascript package already shows this is fine with how the `errors.js`, `es6.js`, etc are already handled
3. We export the requiring type checking exports bundled with their main exports instead of as a separate configuration.
   - I have not seen any instance of a project opting-out of type checking so lets just make it the default.
   - We would keep the `requiring-type-checking.js` files as separate files in the repo since it requires different project settings but we will merge it into the index like we do for `errors.js`, `es6.js`, etc
3. We re-export underlying NI configurations from higher-level composed exports.
   -  Our higher-level libraries already have peerDependencies on the lower-level libraries that projects should not conflict with, so we can just re-export those and let projects use them directly. It's a convenience to avoid projects needing to manage the dependencies and versions of those libraries since they don't control it anyway without conflicting with the higher-level package.

So instead of the proposal of this PR:

```ts
import javascript from '@ni/eslint-config-javascript';
import typescript from '@ni/eslint-config-typescript';
import typescriptRequiringTypeChecking from '@ni/eslint-config-typescript/requiring-type-checking';
import angular from '@ni/eslint-config-angular';
import angularTemplate from '@ni/eslint-config-angular/template';
import angularRequiringTypeChecking from '@ni/eslint-config-angular/requiring-type-checking';
import playwright from '@ni/eslint-config-playwright';
import playwrightRequiringTypeChecking from '@ni/eslint-config-playwright/requiring-type-checking';
```

I am proposing:

```ts
import { javascript } from '@ni/eslint-config-javascript';
import { javascript, typescript } from '@ni/eslint-config-typescript';
import { javascript, typescript, angular. angularTemplate } from '@ni/eslint-config-angular';
import { javascript, typescript, playwright } from '@ni/eslint-config-playwright';
```

Thoughts on the proposal @jattasNI @TrevorKarjanis?

README.md
```diff
@@ -326,11 +350,11 @@ This option can be adapted for npm scripts, for example.
...
 
 ### Angular linting performance
 
-Deviations from the `@angular-eslint schematic`, `@ni/eslint-config-angular`, and the [`parserOptions.project`](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/parser#parseroptionsproject) configurations can result in significant performance degredation. Fully manual configuration [is not recommended](https://github.com/angular-eslint/angular-eslint#going-fully-manual-not-recommended). Read `@angular-eslint`'s [section on performance](https://github.com/angular-eslint/angular-eslint#eslint-configs-and-performance) for information on addressing slow linting processes.
+Deviations from the `angular-eslint`, `@ni/eslint-config-angular`, and the [`parserOptions.project`](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/parser#parseroptionsproject) configurations can result in significant performance degredation. Read `angular-eslint`'s [section on performance](https://github.com/angular-eslint/angular-eslint/blob/v19.0.0/docs/RULES_REQUIRING_TYPE_INFORMATION.md) for information on addressing slow linting processes.
 
 ## License
```
rajsite, line 359: We should add a section to the README that is detailed migration instructions to the flat configuration we can reference from the change file comments and link to from elsewhere.

packages/eslint-config-javascript/rules/style.js
gokulprasanth-ni: Updated

packages/eslint-config-playwright/index.js
gokulprasanth-ni: Yes, I have confirmed this in the `eslint-plugin-playwright` repository. Both configurations are identical in terms of rules—the only difference is the config format, as shown in the image below.

image

image

packages/eslint-config-javascript/rules/imports.js
TrevorKarjanis: This applies to `rules/es6.js` as well.

change/@ni-eslint-config-angular-2b894e10-2f6e-4ff2-ade2-fae0335c38d8.json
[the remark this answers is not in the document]
jattasNI: Is there a way to error instead of warn for unused disable comments? That would be great to take advantage of: we generally avoid warnings because people tend to ignore them. If there is, could you create an issue to remind us to turn on that setting in a follow up PR?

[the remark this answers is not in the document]
jattasNI: I'm glad to hear it's not too expensive to update each app. What's your latest thinking on [the thread about who will do that work](https://dev.azure.com/ni/DevCentral/_workitems/edit/3053642#10236831): your team or the owning teams?

packages/eslint-config-playwright/index.js
jattasNI: Sounds good, we can leave this as-is.

packages/eslint-config-typescript/lib/extensions-requiring-type-checking.js
jattasNI: Sounds like it has identical functionality so the name change should be innocuous, this seems like a good change.

README.md
```diff
@@ -69,80 +69,102 @@ After installing the lint configuration packages, follow the configuration instr
...
 ### JavaScript configuration
 
-Extend `@ni/eslint-config-javascript` in the [ESLint configuration](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats).
+Spread `@ni/eslint-config-javascript` configurations in your [ESLint flat configuration](https://eslint.org/docs/latest/use/configure/configuration-files-new) (`eslint.config.js`):
 
 ```js
-{
-    extends: '@ni/eslint-config-javascript'
-}
+import javascript from '@ni/eslint-config-javascript';
```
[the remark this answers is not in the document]
jattasNI: I think that would break the pattern proposed in the other thread about naming https://github.com/ni/javascript-styleguide/pull/170#discussion_r2356911759 which relies on being able to export multiple items with distinct names from each library. If we went with `configurations` it wouldn't leave room for different named exports whose names reflect their differences.

If you have a counterproposal given that concern, let's move the conversation to that other thread to keep it all in one place (or chat about it at our meeting this afternoon).

change/@ni-eslint-config-angular-2b894e10-2f6e-4ff2-ade2-fae0335c38d8.json
rajsite: Synced with @jattasNI and @TrevorKarjanis and updated the names in the proposal. Should be good to go 👍

change/@ni-eslint-config-angular-2b894e10-2f6e-4ff2-ade2-fae0335c38d8.json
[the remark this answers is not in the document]
gokulprasanth-ni: We are planning to take up the migration for the apps that does not introduce any major updates. If it is a major update we are planning to raise bugs to the respective squads to take that up.

change/@ni-eslint-config-angular-2b894e10-2f6e-4ff2-ade2-fae0335c38d8.json
gokulprasanth-ni: Updated the code as suggested

README.md
gokulprasanth-ni: Added a section in README for migration related details.

packages/eslint-config-javascript/eslint.config.js
jattasNI: I took a pass at cleaning this up.
1. for `no-extraneous-dependencies` I was able to fix all the issues by either adding an explicit peerDependency in package.json or changing the import to be relative (in the case of the Angular package). https://github.com/ni/javascript-styleguide/pull/170/commits/8d0101324c65616d33145bb4b7d08888ef5e1c1e
2. for `no-unresolved` I found some links that describe this as a known issue so I narrowed the suppression to target just the typescript-eslint package https://github.com/ni/javascript-styleguide/pull/170/commits/f22f19ffe9ff8a0c4feabeb710aa4ad913a26ad6
3. I left `extensions` suppressed since our rule instructions say that's something projects might want to do

README.md
```diff
@@ -276,6 +293,46 @@ The rule configuration files in this package (`index.js`, `typescript.js`, etc)
 
 ESLint offers several [ways to disable a rule for a line or file](https://eslint.org/docs/user-guide/configuring/rules#disabling-rules). Suppressions should be as targeted as possible and should include a comment explaining the suppression.
 
+### Migrating to ESLint Flat Configuration Format
```
jattasNI, line 296: I think we're quite close with this PR. My main remaining concern is validating the impact on some real-world client projects. I'd like to see proof-of-concept migrations of one or two projects (e.g. a SystemLinkShared library and an SLE app). You mention in the PR description that you had this working for one app. Can you push that branch into a draft PR? And do the basic setup in another project like SLShared? You don't have to get the build totally passing, I just want to get a sense for the changes required and make sure we're not setting ourselves up for a huge amount of follow up work.

packages/eslint-config-angular/index.js
```diff
@@ -1,9 +1,9 @@
-import typescript from '@ni/eslint-config-typescript';
+import { typescriptConfig } from '@ni/eslint-config-typescript';
 import angular from 'angular-eslint';
 
-export default [
+export const angularConfig = [
```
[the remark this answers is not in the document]
gokulprasanth-ni: The named exports were updated based on [this ](https://github.com/ni/javascript-styleguide/pull/170#discussion_r2356911759)comment. And the `eslint.config.js` still uses a default export because it should have default export as per the migration guide. Refer to [this ](https://eslint.org/docs/latest/use/configure/configuration-files)for more info.

packages/eslint-config-javascript/eslint.config.js
```diff
@@ -1,13 +1,14 @@
...
-        files: ['**/*.js'],
-        rules: {
-            'import/no-default-export': 'off',
-            'import/extensions': 'off',
-            'import/no-extraneous-dependencies': 'off',
-        },
-    },
+	...javascriptConfig,
+	{
+		ignores: ['eslint.config.js'],
```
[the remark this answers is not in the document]
gokulprasanth-ni: Updated as per suggestion

packages/eslint-config-typescript/eslint.config.js
```diff
@@ -1,14 +1,24 @@
-import javascript from '@ni/eslint-config-javascript';
+import { javascriptConfig } from '@ni/eslint-config-javascript';
 
 export default [
-    ...javascript,
+    ...javascriptConfig,
+    {
+        ignores: ['eslint.config.js'],
```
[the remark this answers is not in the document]
gokulprasanth-ni: Updated as per suggestion

packages/eslint-config-angular/eslint.config.js
```diff
@@ -1,13 +1,14 @@
...
-        files: ['**/*.js'],
-        rules: {
-            'import/no-default-export': 'off',
-            'import/no-extraneous-dependencies': 'off',
-            'import/no-unresolved': 'off',
-        },
-    },
+	...javascriptConfig,
+	{
+		ignores: ['eslint.config.js'],
```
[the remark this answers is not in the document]
gokulprasanth-ni: Updated as per suggestion

tests/playwright/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+	"files": ["./index.ts"],
+	"compilerOptions": {
+		"strictNullChecks": true,
+		"moduleResolution": "node",
```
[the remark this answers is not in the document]
gokulprasanth-ni: Yes, It should be node16. Updated the same.

tests/print-evaluated-rules/eslint.config.js
```diff
@@ -1,12 +1,8 @@
...
-    ...javascript,
+    ...javascriptConfig,
     {
-        files: ['*.js'],
-        rules: {
-            'import/no-default-export': 'off',
-            'import/no-extraneous-dependencies': 'off',
-        },
-    },
+        ignores: ['eslint.config.js']
```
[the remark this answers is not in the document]
gokulprasanth-ni: Updated as per suggestion

README.md
```diff
@@ -276,6 +293,46 @@ The rule configuration files in this package (`index.js`, `typescript.js`, etc)
...
+
```
jattasNI, line 297: Some more notes to include in this section:
1. mention the `@stylistic` rules that changed names. Clients may need to modify rule configurations in their config file or rule suppressions in their source to use the new names.
2. mention the workaround you mentioned at UI WG about needing to explicitly link to the config file from angular.json
3. mention the strictNullChecks change. Clients with strictNullChecks enabled can remove rules they had previously overridden. Clients with strictNullChecks disabled may need to add configurations to maintain previous behavior
4. In the PR change files, link to this section instead of mentioning just the strictNullChecks part of the change.

jattasNI (approved):

README.md
```diff
@@ -333,6 +333,44 @@ Instead of using the `extends` property, import the configuration packages you n
...
 
+5. **@stylistic rules**
+   The following rules are moved from `@typescript-eslint` to `@stylistic`. Update any project config overrides and inline suppressions to the new names:  
+   - `@typescript-eslint/member-delimiter-style` → `@stylistic/member-delimiter-style`  
+   - `@typescript-eslint/type-annotation-spacing` → `@stylistic/type-annotation-spacing`  
+
+   Other deprecated rules may also now reside under `@stylistic`. Search your codebase for old suppressions (e.g. `eslint-disable @typescript-eslint/...`) and rename them as needed.
+
+6. **Angular CLI linkage**
+   When using `eslint.config.js` with Angular, explicitly point the Angular workspace at the flat config. In `angular.json` (per project):
```
TrevorKarjanis, line 344: Consider rephrasing to the following which is easier to understand, "Configure the linter in `angular.json` for each project in Angular workspaces to use the `eslint.config.js` ESLint configuration. <example>"

README.md
```diff
@@ -333,6 +333,44 @@ Instead of using the `extends` property, import the configuration packages you n
...
+             "eslintConfig": "./eslint.config.js"
+           }
+         }
+       }
+     }
+   }
+   ```
+
+7. **strictNullChecks** is now enabled by default.  
+   - If your project has `strictNullChecks` enabled, remove previously overridden strictNullChecks rules.
```
TrevorKarjanis, line 362: Rather than using the "strictNullChecks" option as a word, consider a phrase like the following.
> - If your project has `strictNullChecks` enabled, remove previously overridden strict rules for null checks.
> - If your project has `strictNullChecks` disabled, disable strict rules for null checks to maintain previous behavior.

TrevorKarjanis (approved):

change/@ni-eslint-config-angular-2b894e10-2f6e-4ff2-ade2-fae0335c38d8.json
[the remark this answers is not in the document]
TrevorKarjanis: Yes, I'd like to get this in for the Node.js upgrade I am working on..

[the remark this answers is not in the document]
rajsite: Spent a significant amount of time reviewing today but unfortunately didn't finish. I updated our configs to use the [`defineConfig` helper shown throughout the eslint docs](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file), updated the exports pattern to use the single top-level entrypoint pattern, updated the lint files to lint local files correctly, updated patterns to use examples from the eslint docs more consistently (avoid spread operators on all the configs unnecessarily, avoiding duplicating configuration in config files).

A few things I'm still investigating are the regression in checks for inline templates ([notice how the angular test was updated to manually specify overrides for inline template](https://github.com/ni/javascript-styleguide/pull/170/files#diff-421ed92b4fcaf9511a2977573a57b56861647b02a3af58004a420cf00fcc289dR39) when it previously didn't, this is a regression).

I'm hoping to get some more time to review / investigate the angular inline template regression tomorrow but I'm not 100% sure. Maybe @gokulprasanth-ni can provide insight why that rule is duplicated into the test when it is set in the template ruleset.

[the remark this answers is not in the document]
rajsite: Makes sense and thanks for pushing the update! 👍 Also pushed some README updates and I think it's good to go

.github/CODEOWNERS
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2023 National Instruments
+# Copyright (c) 2025 National Instruments
```
rajsite, line 1: @jattasNI I pushed some significant updates so reset you to make the last sanity pass and then free to release

Offline asked about validation of our playwright packages which have extremely old dependencies and I'm concerned about the eslint flat config support

.github/CODEOWNERS
```diff
@@ -1,4 +1,4 @@
...
```
rajsite, line 2: Actually thought of one more thing, I'll add cjs tests because that should be a smaller change for some projects that cannot switch to `type: module` easily yet (nimble being one of those). Eslint Flat config should support both (lists `eslint.config.cjs` as supported) but worthwhile having smoke tests to ensure our plugins are behaving well.

rajsite: Tested manually in each of the smoke tests in `tests/` that removing `"type":"module"` from the `package.json` and rewriting as commonjs, ie

going from esm:

```js
import { defineConfig } from 'eslint/config';
import { javascriptConfig } from '@ni/eslint-config-javascript';
import { typescriptConfig } from '@ni/eslint-config-typescript';

export default defineConfig([
    javascriptConfig,
    {
        files: ['**/*.ts'],
        extends: typescriptConfig,
        languageOptions: {
            parserOptions: {
                project: ['./tsconfig.json'],
                tsConfigRootDir: import.meta.dirname,
            },
        },
    },
]);
```

to commonjs:

```js
const { defineConfig } = require('eslint/config');
const { javascriptConfig } = require('@ni/eslint-config-javascript');
const { typescriptConfig } = require('@ni/eslint-config-typescript');

module.exports = defineConfig([
    javascriptConfig,
    {
        files: ['**/*.ts'],
        extends: typescriptConfig,
        languageOptions: {
            parserOptions: {
                project: ['./tsconfig.json'],
                tsConfigRootDir: __dirname,
            },
        },
    },
]);
```

Works as intended. It does make an annoying warning about experimental cjs support in node v22:
```
C:\dev\github\ni\javascript-styleguide>npm run test -w tests\angular    

> @ni/angular-test@1.0.0 test
> eslint .

(node:72356) ExperimentalWarning: CommonJS module C:\dev\github\ni\javascript-styleguide\tests\angular\eslint.config.js is loading ES Module C:\dev\github\ni\javascript-styleguide\packages\eslint-config-javascript\index.js using require().
Support for loading ES Module in require() is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
```
 but that's good enough for me and doesn't warrant checking in additional test configurations. Closing.

.github/CODEOWNERS
TrevorKarjanis: I'm here. I'm here sheesh :roll_eyes:

TrevorKarjanis (approved):

rajsite:
@gokulprasanth-ni please note some of the changes in recent updates we'd want to see in other PRs downstream when updating reops to use the config such as:
- Using the `defineConfig` pattern in all `eslint.config.js` files
- Using `tsConfigRootDir: import.meta.dirname,` instead of doing path parsing
- Avoiding unnecessary array spreads on configuration, i.e. just need `extends: typescriptConfig,`, etc
- For non-trivial transforms to `"type": "module"` those changes should be staged separately (a pr to transition to module first and then eslint chnages) or eslint should use [commonjs style configuration](https://github.com/ni/javascript-styleguide/pull/170#discussion_r2392464487) and file tech debt to transition the package to module

Issue #176, opened 2025-11-25
Evaluate the project, projectService, and root workspace configuration requirement for Angular


TrevorKarjanis:
**Describe the issue**
Originally [#173](https://github.com/ni/javascript-styleguide/pull/173#discussion_r2424952842)

> I want to do a further evaluation of the necessity for the `project` configuration and the effect of only using the root tsconfig.json. I expect this mostly includes the configurations not accounted for in the other TS configuration files, e.g. the application, test specs, and library files.

> We'll need to upstream identified issues and recommendations to [angular/eslint](https://github.com/angular-eslint/angular-eslint), likely in the form of a bug report. There is some existing commentary about the `projectService` configuration (issues https://github.com/angular-eslint/angular-eslint/pull/1872#discussion_r1713855327 and https://github.com/angular-eslint/angular-eslint/issues/609#issuecomment-2907678163).

> Additionally, Ravi informed me that a root workspace configuration is required even if the workspace is made up only of projects. This should be evaluated as well.

At least the following can be completed and information can be documented in this bug report.

1. Generate a new workspace with the latest version of Angular.
    - This workspace will have one root project.
3. Install and run the schematic for angular-eslint.
4. Document any notable characteristics of the default generated configuration.
5. Apply the rest of this style guide as documented in the README.
6. Document notable effects of using the `projectService` configuration with a root and child projects.
    - Consider evaluating the tool source to develop expectations for the identification of `tsconfig` files with `projectService`.
    - Which TS configurations are used for the root project?
    - Which TS configurations are used for the child projects?
    - Are the TS configurations for tests used?
7. Document notable effects of using the `project` configuration.
8. Attempt to configure the workspace without a root configuration and the document the results.

PR #181, opened 2025-11-26
Test with an Angular workspace in the Angular smoke test.

### LINKED ISSUE #177: Update the tests for Angular to run on an Angular workspace
**Describe the issue**
Originally [#173](https://github.com/ni/javascript-styleguide/pull/173)

> The tests for Angular also need to be expanded to validate that our configuration and how we document its use works with Angular workspaces and the VSCode extensions. This will be done in a separate PR.

The new flat configuration has characteristics that are specific to Angular and distinct between a root application and child projects.

1. Replace the `tests/angular` index file with an Angular workspace that includes a root and child project.
3. Ensure the file patterns in `angular.json` include JavaScript files.
4. Ensure the child project uses the `project` configuration to reference its TS configuration.
5. Develop a smoke test that ensures the linter configuration applies with the correct linter configurations with the root and separately child project.

TrevorKarjanis:
Resolves #177. Test with an Angular workspace to validate that the style guide configuration works with the workspace configuration, root applications, libraries, language and parser options, ng lint, and the standalone eslint command. This also validates the instructions provided in the README. The original ignore custom attributes test has been integrated into the root application component template. Additionally, there is a root component test with HTML file and library component test with inline template.

The workspace has been reduced to the necessities of a test which satisfies the criteria.

- Test with an Angular workspace in the Angular smoke test.

Testing
- Verified that the automated tests pass (npm test)
  - ng lint
  - eslint
  - cd projects/angular-library && eslint .
- Manually verified that the application builds and starts


tests/angular/package.json
```diff
@@ -1,20 +1,45 @@
...
+    "rxjs": "~7.8.0",
+    "tslib": "^2.3.0",
+    "zone.js": "~0.14.10"
   },
   "devDependencies": {
-    "@types/jasmine": "^5.1.4"
+    "@angular-devkit/build-angular": "^18.2.21",
+    "@angular/cli": "^18.2.21",
+    "@angular/compiler-cli": "^18.2.0",
+    "@angular/localize": "^18.2.14",
```
rajsite: We are imminently working on Angular 19 upgrade, thoughts on just jumping straight to that?

tests/angular/package.json
```diff
@@ -1,20 +1,45 @@
...
-  },
-  "peerDependencies": {
-    "@ni/eslint-config-angular": "*",
-    "@ni/eslint-config-javascript": "*"
+    "start": "ng serve",
+    "build": "ng build && ng build angular-library",
+    "lint": "ng lint",
+    "ng-test": "ng test --watch=false",
+    "watch": "ng build --watch --configuration development",
+    "test": "ng lint && eslint . && cd projects/angular-library && eslint ."
```
rajsite: maybe we should also go ahead and run `ng-test` as part of `test` so it gets exercised.

tests/angular/package.json
```diff
@@ -1,20 +1,44 @@
...
+    "start": "ng serve",
+    "build": "ng build && ng build angular-library",
+    "lint": "ng lint",
+    "test": "ng lint && ng test",
+    "watch": "ng build --watch --configuration development"
   },
+  "private": true,
   "dependencies": {
-    "@angular/core": "^18.2.13"
+    "@angular/animations": "^18.2.0",
```
rajsite: Since `ng test` is running on CI it is now currently relying on whatever version of Chrome happens to be on the build agent. We generally would use [playwright](https://github.com/ni/nimble/blob/265570a90abc2dd3cacfcb84383cf0486150b8a4/packages/nimble-components/karma.conf.js#L11) or [puppeteer](https://github.com/karma-runner/karma-chrome-launcher?tab=readme-ov-file#headless-chromium-with-puppeteer) to bring in a versioned copy of Chrome for reproducibility so I'll leave the comment if you're up for it, but the repo is so small scoped / test suite so narrow that I don't feel particularly concerned about the build reproducibility here and am okay either way.

rajsite (approved):

tests/angular/package.json
TrevorKarjanis: I'm not going to worry about that right now. I'm already working on #176, and introducing that capability would derail my current direction quite a bit. Like yourself, I'm not concerned.

tests/angular/README.md
```diff
@@ -1,5 +1,8 @@
 # Angular Smoke Test
 
-## Overview
+This package is a smoke test for using the Angular rules in an Angular workspace. See [CONTRIBUTING.md](../../CONTRIBUTING.md#smoke-tests) for more information on smoke tests.
 
-This package is a smoke test for using the Angular rules in a project. See [CONTRIBUTING.md](../../CONTRIBUTING.md#smoke-tests) for more information.
+## Testing
+
+- The `npm test` script runs the linter to test that the lint configuration functions correctly with `ng lint` and `eslint`.
```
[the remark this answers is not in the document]
TrevorKarjanis: Thanks, moving without thinking.

PR #196, opened 2026-05-11
Add Ok fv attribute ignores to angular eslint

rajsite:
- Added a list of attribute ignores for ok fv components, see https://github.com/ni/nimble/pull/2954
- Update CODEOWNERS to copy the pattern in nimble for fv ownership

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,3 +5,9 @@
 
 # These owners will be the default owners for everything in the repo.
 * @jattasNI @rajsite
+
+# Specific content owners
+/**/fv/** @jattasNI @fredvisser
+
+# Change files don't need explicit reviewers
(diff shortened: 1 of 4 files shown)


.github/CODEOWNERS
```diff
@@ -5,3 +5,9 @@
```
rajsite, line 5: Started some offline discussion with @jattasNI if we want to track this in styleguide packages.

rajsite: Related, we should probably have the selectors be narrowly scoped, i.e. `ok-fv-card[interaction-mode]`

rajsite: brainstorming, other potential locations could be an `@ni-labs` package in this repo or in a separate repo

PR #198, opened 2026-08-06
Exempt pinLocation from i18n rule

holzealy:
The `sl-table-column-mapping` component has a `pinLocation` attribute that does not require translation, so should be included in the list of exemptions.


packages/eslint-config-angular/lib/template-options.js
jattasNI, line 69: Strange, I'm not sure why the Nimble attribute wasn't giving you problems too. Maybe that directory isn't being linted. But let's add this just to be safe.
```suggestion
        'pin-location',
        'selection-mode',
```

jattasNI (approved):

change/@ni-eslint-config-angular-a4278db8-6d43-4e27-90d5-eed1acdc8d6d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add 'pinLocation' i18n exception for sl-table mapping column",
```
jattasNI: For some reason the build isn't triggering, I'm pushing this change to try to kick it into action.
```suggestion
  "comment": "Add i18n exceptions for sl-table and nimble-table mapping column pin location",
```

